PR #35, opened 2025-06-02
Refactor/move plugin package up

lchoquel:
- Moved “plugin” to the same level as “cogt” within **pipelex** for better visibility
- Major cleanup in the unit tests: hierarchy largely flattened
- Added **pytest-xdist** to run unit tests in parallel on multiple CPUs. Not yet integrated into the Makefile, so run it manually with `pytest -n auto` or `pytest -n auto -m "llm"`
- Added `make test-quiet` (shorthand `tq`) to Makefile to run tests without capturing outputs (i.e. without pytest `-s` option)
- Stopped using Fixtures for `pipe_router` and `content_generator`: we're now always getting the singleton from `pipelex.hub`


thomashebrard (approved):
good

antoinepipelex (approved):
lgtm

PR #45, opened 2025-06-03
Refactor/rename pipeline

lchoquel:
### 📝 Description

- rename Mission to Pipeline
- enable to start a pipeline and let in run in the background, getting it's run id, but not waiting for the output
- Makefile: avoid defaulting pytest to verbose. Setup target `make test-xdist` = Run unit tests with xdist, make it the default for shorthand `make t`. The old make t is now make tp (test-with-prints)

### 🔄 Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [X] 🧹 Code refactor
- [ ] ✅ Test update


pipelex/pipeline/pipeline_models.py
```diff
@@ -1,5 +1,5 @@
 from pipelex.types import StrEnum
 
 
-class SpecialMissionId(StrEnum):
+class SpecialPipelineId(StrEnum):
```
thomashebrard, line 4: `Notable`instead of `Special`?

pipelex/pipeline/pipeline.py
```diff
@@ -0,0 +1,5 @@
+from pydantic import BaseModel
+
+
+class Pipeline(BaseModel):
+    pipeline_run_id: str
```
thomashebrard, line 5: do we really need "pipeline" here ? 
I am trying to imagine this code:
```
my_pipeline = Pipeline()
my_pipeline.pipeline_run_id
```
seems heavy

thomashebrard (approved):
seems great

pipelex/pipeline/pipeline_models.py
lchoquel: not better

pipelex/pipeline/pipeline.py
lchoquel: not strictly but for a while, while we transition from the run_pipe API

PR #66, opened 2025-06-06
Refactor/inference methods wrapping and more

lchoquel:
### 📝 Description

- New way (simpler and way more elegant) to wrap inference worker methods (llm, image gen, ocr): instead of the job_func decorators, it uses a base method that calls a private abstract method implemented by each worker, and does stuff before and after
- Introduce new pytest marker pipelex_api, applied to client methods instead of inference, llm etc.
- Added target `make test-pipelex-api` aka `make ta` to run just those tests.
- Fix error regarding pipe stack and PipeParallel

### 🔄 Type of Change
- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [X] 🧹 Code refactor
- [X] ✅ Test update


pipelex/cogt/llm/llm_worker_abstract.py
```diff
@@ -71,17 +57,98 @@ def check_can_perform_job(self, llm_job: LLMJob, func_name: LLMWorkerJobFuncName
...
+                f"LLM Worker error: Instructor failed after retry with llm '{self.llm_engine.tag}': {exc}\nLLMPrompt: {llm_job.llm_prompt.desc}"
+            ) from exc
+
+        # Cleanup result
+        if hasattr(result, "_raw_response"):
+            delattr(result, "_raw_response")
+
+        # Report job
+        llm_job.llm_job_after_complete()
+        if self.reporting_delegate:
```
thomashebrard, line 143: Great! I believe this function could me wrapped and used in the gen_text as well!

thomashebrard (approved):
ok

pipelex/cogt/llm/llm_worker_abstract.py
lchoquel: that is indeed the case

PR #67, opened 2025-06-09
Fix/input requirements with attribute paths

lchoquel:
### 🔄 Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [X] 🧹 Code refactor
- [X] ✅ Test update


pipelex/pipe_operators/pipe_img_gen.py
```diff
@@ -55,6 +58,88 @@ class PipeImgGen(PipeOperator):
     seed: Optional[Union[int, Literal["auto"]]] = None
     output_multiplicity: PipeOutputMultiplicity
 
+    img_gen_prompt_var_name: Optional[str] = None
+
+    @field_validator("img_gen_prompt_var_name")
```
thomashebrard, line 63: sometimes we have imgg, sometimes img_gen. We should decide on smth, or have I missed smth?

pipelex/pipe_operators/pipe_img_gen.py
```diff
@@ -55,6 +58,88 @@ class PipeImgGen(PipeOperator):
...
+            if concept_provider.is_compatible_by_concept_code(
+                tested_concept_code=input_concept_code,
+                wanted_concept_code=NativeConcept.TEXT.code,
+            ):
+                self.img_gen_prompt_var_name = input_name
+                candidate_prompt_var_names.append(input_name)
+            else:
+                inadequate_input_concept_error = StaticValidationError(
+                    error_type=StaticValidationErrorType.INADEQUATE_INPUT_CONCEPT,
+                    domain_code=self.domain,
```
thomashebrard, line 100: Why this mecanic is not in PipeOperation?

pipelex/pipe_operators/pipe_img_gen_factory.py
```diff
@@ -52,11 +51,10 @@ def make_pipe_from_blueprint(
             domain=domain_code,
             code=pipe_code,
             definition=pipe_blueprint.definition,
-            input_concept_code=pipe_blueprint.input,
+            inputs=PipeInputSpec(root=pipe_blueprint.inputs or {}),
```
thomashebrard, line 54: Create a capital_letter global variable for "{}"

thomashebrard (approved):
Super nice

pipelex/pipe_operators/pipe_img_gen.py
lchoquel: partial cleanup, needs fini refactor, yes

pipelex/pipe_operators/pipe_img_gen.py
lchoquel: because the validation is very specific to each pipe operation, because each has some implicit rules: OCR needs an image or pdf as input, image generation needs a prompt either as attribute or as input and it must be a single inout etc.

pipelex/pipe_operators/pipe_img_gen_factory.py
lchoquel: to initialize an empty dict? that would be going too far

PR #118, opened 2025-06-26
Feature/storage provider ingestion

Abdujabbar:
### 📝 Description

Added abstract class for storage providers, ingestion functions for pipelex hub.

### 🔄 Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [X] 📚 Documentation update
- [ ] 🧹 Code refactor
- [ ] ✅ Test update

### 📋 Checklist

- [X] Linting / type-checking / unit tests pass locally with my changes
- [X] My code follows the style guidelines of this project
- [X] I've made corresponding changes to the documentation
- [X] My changes generate no new warnings

### DIFF
--- a/pipelex/tools/storage/storage_provider_abstract.py
+++ b/pipelex/tools/storage/storage_provider_abstract.py
@@ -1,13 +1,11 @@
 from abc import ABC, abstractmethod
 
-from botocore.client import BaseClient
-
 
 class StorageProviderAbstract(ABC):
     @abstractmethod
-    def get_object(self, bucket_name: str, object_key: str) -> bytes:
+    def load(self, uri: str) -> bytes:
         pass
 
     @abstractmethod
-    def get_client(self) -> BaseClient:
+    def store(self, data: bytes) -> str:
         pass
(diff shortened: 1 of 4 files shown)


pipelex/tools/storage/storage_provider_abstract.py
lchoquel, line 1: The StorageProviderAbstract should not be related to aws or boto3: leave it to the concrete subclass to use an S3 client. The abstract interface should just be to save/load:
`load(uri: str) -> bytes`
and
`store(data: bytes) -> uri`

lchoquel (changes requested):
Please make the StorageProviderAbstract independent of the storage provider

pipelex/tools/storage/storage_provider_abstract.py
Abdujabbar: @lchoquel  Thanks for the feedback! 

I've changed the abstract class, now StorageProviderAbstract does not depend on s3 or boto3, and I also added the abstract function load/store, expecting that concrete classes will implement them.

lchoquel (approved):
OK

PR #129, opened 2025-06-29
Refactor/plugin management

lchoquel:
### 📝 Description
- handle external plugins for LLMs

### 🔄 Type of Change
- [ ] 🐛 Bug fix
- [X] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [X] 🧹 Code refactor
- [ ] ✅ Test update

### 🧪 Tests
- test_external_plugin.py


pipelex/cogt/llm/llm_models/llm_model_library.py
```diff
@@ -100,6 +100,54 @@ def get_llm_model(
...
+        llm_platform_choice: LLMPlatformChoice,
+    ) -> Optional[LLMModel]:
+        found_llm_models = [llm_model for llm_model in self.root if llm_model.llm_name == llm_name]
+        if not found_llm_models:
+            return None
+        elif len(found_llm_models) == 1:
+            # only one so it is the latest we've got
+            llm_model = found_llm_models[0]
+        elif chosen_model := next((model for model in found_llm_models if model.version == LATEST_VERSION_NAME), None):
+            # we've got a version named 'latest' so we use it
```
thomashebrard: This is double code, it can be simplified

pipelex/cogt/llm/llm_worker_internal_abstract.py
```diff
@@ -0,0 +1,67 @@
...
+
+from pipelex.cogt.exceptions import LLMCapabilityError
+from pipelex.cogt.llm.llm_job import LLMJob
+from pipelex.cogt.llm.llm_models.llm_engine import LLMEngine
+from pipelex.cogt.llm.llm_worker_abstract import LLMWorkerAbstract
+from pipelex.cogt.llm.structured_output import StructureMethod
+from pipelex.reporting.reporting_protocol import ReportingProtocol
+
+
+class LLMWorkerInternalAbstract(LLMWorkerAbstract):
```
thomashebrard, line 13: Why is the reason for calling it "internal"?

pipelex/cogt/llm/llm_models/llm_model_library.py
lchoquel: yes, I forgot to get back to it

pipelex/cogt/llm/llm_worker_internal_abstract.py
lchoquel: Our LLM Workers are plugins.
In general, plugins come from outside, but we only had internal home-made plugins.

Now, this PR makes it easy to plug a plugin from outside. Coding it made clear that part of our LLMWorkerAbstract was related to the way we handle our internal LLM workers and should not apply to plugins from outside. This is now reflected in the separation of concern between `LLMWorkerAbstract` (any plugin) and `LLMWorkerInternalAbstract` (only our internal plugins)

PR #188, opened 2025-07-22
Feature/implicit memory

lchoquel:
### 📝 Description

- Implicit memory to call execute_pipeline with a very simple input_memory
- changed InputRequirement to support more details required about inputs, not only concept_code, created subclasses useful at different stages of analysis, cleaner than having anonymous tuple definitions: InputRequirement/NamedInputRequirement/TypedNamedInputRequirement


pipelex/cli/_cli.py
```diff
@@ -193,7 +219,8 @@ def show_pipe(
         typer.Argument(help="Pipeline code to show definition for"),
     ],
     relative_config_folder_path: Annotated[
-        str, typer.Option("--config-folder-path", "-c", help="Relative path to the config folder path")
+        str,
+        typer.Option("--config-folder-path", "-c", help="Relative path to the config folder path"),
     ] = "./pipelex_libraries",
```
thomashebrard, line 224: Why the formatting is different ?

thomashebrard (approved):

pipelex/core/concept_library.py
```diff
@@ -178,3 +179,12 @@ def is_image_concept(self, concept_code: str) -> bool:
         is_image_class = bool(pydantic_model and issubclass(pydantic_model, ImageContent))
         refines_image = self.is_compatible_by_concept_code(tested_concept_code=concept.code, wanted_concept_code=NativeConcept.IMAGE.code)
         return is_image_class or refines_image
+
+    @override
+    def search_for_concept_in_domains(self, concept_name: str, search_domains: List[str]) -> Optional[Concept]:
+        for domain in search_domains:
+            concept_code = ConceptCodeFactory.make_concept_code(domain=domain, code=concept_name)
+            if found_concept := self.get_concept(concept_code=concept_code):
+                return found_concept
```
thomashebrard, line 188: Why not do self.root[domain][concept] annd catch the KeyError?

pipelex/cli/_cli.py
lchoquel: added comma I think the fromatting rules applied in Cursor are not the same when I edit a file thru a different repo, lilke if I open the Pipelex file from inside my Cookbook workspace

pipelex/core/concept_library.py
lchoquel: root is underlying implementation of ConceptLibrary. Even though we are in ConceptLibrary and could use underlying features, it's safer to use get_concept() = the public accessor and part of the ConceptProviderAbstract interface. Because if we invent a new way to get a concept… it won't need to change a thing

thomashebrard (approved):

PR #236, opened 2025-08-24
Refactor/concept

thomashebrard:
### 🔄 Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🧹 Code refactor
- [ ] ✅ Test update

### 📋 Checklist


pipelex/cogt/content_generation/jinja2_generate.py
lchoquel, line 1: These changes are not needed, the bug you faced is fixed by other means and anyway it's not consistent with our architecture to process the template when rendering, it's done upstream on order to identify used variables.
Also, the check_jinja2_parsing's only job is to raise if the template has a syntax error, which the jinja2 rendering will do anyway.

pipelex/core/bundles/pipelex_bundle_blueprint.py
```diff
@@ -43,3 +45,18 @@ class PipelexBundleBlueprint(BaseModel):
...
+
+    @field_validator("domain", mode="before")
+    @classmethod
+    def validate_domain_syntax(cls, domain: str) -> str:
+        DomainBlueprint.validate_domain_code(code=domain)
+        return domain
+
+    @model_validator(mode="before")
+    @classmethod
+    def validate_non_valid_values_in_blueprint(cls, values: Dict[str, Any]) -> Dict[str, Any]:
```
lchoquel: this is just checking we don't have extra keys, right?

the pydantic way to achieve that is to add `model_config = ConfigDict(extra="forbid")`

pipelex/core/concepts/concept.py
```diff
@@ -20,107 +21,77 @@ class Concept(BaseModel):
     domain: str
     definition: str
     structure_class_name: str
-    refines: List[str] = Field(default_factory=list)
+    refines: Optional[str] = None
```
lchoquel, line 22: please add `# TODO: restore possibility of multiple refiles`

pipelex/core/concepts/concept.py
```diff
@@ -20,107 +21,77 @@ class Concept(BaseModel):
...
+    def validate_domain(cls, domain: str) -> str:
+        DomainBlueprint.validate_domain_code(code=domain)
+        return domain
 
-            # Validate the domain and concept syntax for the full code
-            domain, code = cls.extract_domain_and_concept_from_str(concept_str=full_code)
-            cls.validate_concept_code_syntax(code=code, concept_code=full_code, domain_field=domain)
-            cls.validate_domain_syntax(domain=domain, code=full_code, domain_field=domain)
+    @model_validator(mode="after")
+    def validate_concept(self) -> Self:
```
lchoquel: doesn't actually validate anything, remove

pipelex/core/concepts/concept.py
```diff
@@ -20,107 +21,77 @@ class Concept(BaseModel):
...
-        raise ConceptError(f"Could not extract domain and concept from '{concept_str}'")
+    def sentence_from_concept(cls, concept: "Concept") -> str:
+        return pascal_case_to_sentence(name=concept.code)
 
-    @classmethod
-    def extract_concept_name_from_str(cls, concept_str: str) -> str:
-        _, concept = cls.extract_domain_and_concept_from_str(concept_str=concept_str)
-        return concept
+    @property
+    def node_name(self) -> str:
```
lchoquel: not used, remove

pipelex/core/concepts/concept.py
```diff
@@ -20,107 +21,77 @@ class Concept(BaseModel):
...
-        domain, _ = cls.extract_domain_and_concept_from_str(concept_str=concept_str)
-        return domain
+    def is_native_concept(cls, concept: "Concept") -> bool:
+        return ConceptBlueprint.is_native_concept_string_or_concept_code(concept_string_or_concept_code=concept.concept_string)
 
     @classmethod
-    def concept_str_contains_domain(cls, concept_str: str) -> bool:
-        """Check if the concept code contains a domain and is in the form <domain>.<concept_code>"""
-        return "." in concept_str and len(concept_str.split(".")) == 2
+    def are_concept_compatible(cls, concept_1: "Concept", concept_2: "Concept", strict: bool = False) -> bool:
```
lchoquel, line 59: You're testing for compatibility of structure classes. So you could rename this `is_structure_compatible`.
But most of all, you've lost the existing code which tested if concept was compatible semantically by looking up the refines transitively. It is crucial to bring it back.

pipelex/core/concepts/concept.py
```diff
@@ -20,107 +21,77 @@ class Concept(BaseModel):
...
 
-    @classmethod
-    def is_native_concept(cls, concept_str: str) -> bool:
-        if Concept.concept_str_contains_domain(concept_str=concept_str):
-            domain = Concept.extract_domain_from_str(concept_str=concept_str)
-            return domain == SpecialDomain.NATIVE.value
-        else:
-            return concept_str in NativeConcept.names()
+                # Check if concept_1 has a field that is of type concept_2_class
+                return ClassRegistryUtils.has_compatible_field(concept_1_class, concept_2_class)
```
lchoquel, line 83: I don't see how having a concept_2 field will make the structure compatible

pipelex/core/concepts/concept.py
```diff
@@ -132,3 +103,17 @@ def is_valid_structure_class(cls, structure_class_name: str) -> bool:
             if KajsonManager.get_class_registry().has_class(name=structure_class_name):
                 log.warning(f"Concept class '{structure_class_name}' is registered but it's not a subclass of StuffContent")
             return False
+
+    @classmethod
+    def get_structure(cls, class_name: str, structure_blueprint: Dict[str, ConceptStructureBlueprint]) -> str:
```
lchoquel: this name does not make it clear it's actually **generating** a **string of python code**.

same goes for the method that does the job, `generate_from_structure_blueprint` it clear it's actually returning a **string of python code**

pipelex/create/structured_output_generator.py
lchoquel, line 1: Seeing how it's used, I think the args of `generate_from_structure_blueprint` (class_name: str, structure_blueprint: Dict[str, ConceptStructureBlueprint])` should be passed to init, and you'll avoid having to passe them around the private methods

pipelex/core/concepts/concept_blueprint.py
```diff
@@ -29,23 +37,82 @@ class ConceptStructureBlueprint(BaseModel):
...
 
         return self
 
+    def _validate_default_value_type(self) -> None:
+        """Validate that default_value matches the specified type."""
+        if self.type is None or self.default_value is None:
+            return
+
+        # Validate based on the type
+        if self.type == ConceptStructureBlueprintFieldType.TEXT:
```
lchoquel: you must use match/case construct
and list all the cases, including the ones you want to skip  (don't use `case _:`) but you can match multiple values with `|`

pipelex/core/concepts/exceptions.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self, code: str):
+        self.code = code
+        super().__init__(f"Concept code '{self.code}' must be PascalCase (letters and numbers only, starting with uppercase, without `.`)")
+
+
+class ConceptStringError(Exception):
+    def __init__(self, concept_string: str, message: Optional[str] = None):
+        super().__init__(
+            f"Concept string '{concept_string}' is invalid. It should contain a domain in snake_case "
+            "and a concept code in PascalCase separated by one dot."
```
lchoquel: This is not the right place to document the error. you can use the `message` attribute for that.
So that the explanation and the actual validation code will be in the same place in the code.

pipelex/core/concepts/exceptions.py
```diff
@@ -0,0 +1,26 @@
...
+            else message
+        )
+
+
+class ConceptStringOrConceptCodeError(Exception):
+    def __init__(self, concept_string_or_concept_code: str):
+        super().__init__(
+            f"concept_string_or_concept_code '{concept_string_or_concept_code}' is invalid. "
+            "It should either contain a domain in snake_case and a concept code in PascalCase separated by one dot, "
+            "or be a concept code in PascalCase."
```
lchoquel: same as above

pipelex/core/concepts/concept_blueprint.py
```diff
@@ -55,27 +122,111 @@ class ConceptBlueprint(BaseModel):
...
+                    concept_string,
+                    f"Concept code '{concept_code}' is not a native concept, so the domain must be '{SpecialDomain.NATIVE.value}', or nothing.",
+                )
+
+    @field_validator("refines", mode="before")
+    @classmethod
+    def validate_refines(cls, refines: Optional[str] = None) -> Optional[str]:
+        def is_native_concept(concept_ref: str) -> bool:
+            """Check if a concept reference is a native concept (short or fully qualified form)."""
+            native_concept_values = [native_concept.value for native_concept in NativeConceptEnum]
```
lchoquel: make this a classmethod of `NativeConceptEnum`

pipelex/core/concepts/concept_blueprint.py
```diff
@@ -55,27 +122,111 @@ class ConceptBlueprint(BaseModel):
...
+            return False
+
+        if refines is not None:
+            if not is_native_concept(refines):
+                raise ConceptBlueprintError(f"Forbidden to refine a non-native concept: '{refines}'. Refining non-native concepts will come soon.")
+            cls.validate_concept_string_or_concept_code(concept_string_or_concept_code=refines)
+        return refines
+
+    @classmethod
+    def forbiden_having_refines_and_structure(cls, values: Dict[str, Any]) -> Dict[str, Any]:
```
lchoquel: forbidden

pipelex/core/concepts/concept_blueprint.py
```diff
@@ -55,27 +122,111 @@ class ConceptBlueprint(BaseModel):
...
+        if values.get("refines") and values.get("structure"):
```
lchoquel: we have a helper for that: `from pipelex.tools.typing.validation_utils import has_more_than_one_among_attributes_from_list`

pipelex/core/concepts/concept_blueprint.py
```diff
@@ -55,27 +122,111 @@ class ConceptBlueprint(BaseModel):
...
+            raise ConceptBlueprintError(
+                f"Forbidden to have refines and structure at the same time: `{values.get('refines')}` "
+                f"and `{values.get('structure')}` for concept that has the definition `{values.get('definition')}`"
+            )
+        return values
+
+    @model_validator(mode="before")
+    def model_validate_blueprint(cls, values: Dict[str, Any]) -> Dict[str, Any]:
+        if isinstance(values, ConceptBlueprint):
```
lchoquel: I don't understand:
`values: Dict[str, Any]` versus `isinstance(values, ConceptBlueprint)`

pipelex/core/concepts/concept_factory.py
```diff
@@ -1,168 +1,192 @@
...
-        structure_class_name: str
-        refines: List[str]
-        if Concept.concept_str_contains_domain(concept_str=concept_str):
-            concept_name = Concept.extract_concept_name_from_str(concept_str=concept_str)
+    def make_implicit_concept(cls, concept_string: str) -> Concept:
+        ConceptBlueprint.validate_concept_string(concept_string=concept_string)
+        return Concept(
+            code=concept_string.split(".")[1],
+            domain=SpecialDomain.IMPLICIT.value,
+            definition="",
```
lchoquel: The empty `definition` could create issues, for instance if you write a prompt and want to inject concept definition into it.
Easy solution: `definition=code`

pipelex/core/concepts/concept_factory.py
```diff
@@ -1,168 +1,192 @@
...
-    ) -> Concept:
-        structure_class_name: str
-        refines: List[str]
-        if Concept.concept_str_contains_domain(concept_str=concept_str):
-            concept_name = Concept.extract_concept_name_from_str(concept_str=concept_str)
+    def make_implicit_concept(cls, concept_string: str) -> Concept:
+        ConceptBlueprint.validate_concept_string(concept_string=concept_string)
+        return Concept(
+            code=concept_string.split(".")[1],
+            domain=SpecialDomain.IMPLICIT.value,
```
lchoquel: don't need `.value`

pipelex/core/concepts/concept_factory.py
```diff
@@ -1,168 +1,192 @@
...
+        if "." in concept_string_or_concept_code:
+            # Is a concept string.
+            return concept_string_or_concept_code.split(".")
         else:
-            concept_name = concept_str
-        if Concept.is_valid_structure_class(structure_class_name=concept_name):
-            # structure is set implicitly, by the concept's code
-            structure_class_name = concept_name
-            refines = []
+            if concept_string_or_concept_code in [native_concept.value for native_concept in NativeConceptEnum]:  # Is a native concept code
```
lchoquel: code smell: do what your comment suggests, make it a `NativeConceptEnum` method

pipelex/core/concepts/concept_factory.py
```diff
@@ -1,168 +1,192 @@
...
+            code=native_concept_data.code,
+            domain=SpecialDomain.NATIVE,
+            definition=native_concept_data.definition,
+            structure_class_name=native_concept_data.content_class_name,
+        )
+
+    @classmethod
+    def make_domain_and_concept_code_from_concept_string_or_concept_code(
+        cls, domain: str, concept_string_or_concept_code: str, concept_codes_from_the_same_domain: Optional[List[str]] = None
+    ) -> List[str]:
```
lchoquel: What you're returning feels more like a Tuple.
But as Python does not support naming tuple items, let's make this more solid with a BaseModel holding both attributes `domain` and `concept_code`.

Also: keep in mind that in the future, if we manage namespaces, domain does could also have "." in them (for hierarchy of domains). So it would be safer to use `rsplit(".", 1)` (splits once, only at the last dot)

pipelex/core/concepts/concept_factory.py
```diff
@@ -1,168 +1,192 @@
...
-            refines = []
+            if concept_string_or_concept_code in [native_concept.value for native_concept in NativeConceptEnum]:  # Is a native concept code
+                return [SpecialDomain.NATIVE.value, concept_string_or_concept_code]
+
+            elif (
+                concept_codes_from_the_same_domain and concept_string_or_concept_code in concept_codes_from_the_same_domain
+            ):  # Is a concept code from the same domain
+                return [domain, concept_string_or_concept_code]
+            else:
+                return [SpecialDomain.IMPLICIT.value, concept_string_or_concept_code]
```
lchoquel: don't need `.value`

pipelex/core/concepts/concept_factory.py
```diff
@@ -1,168 +1,192 @@
...
+
+    @classmethod
+    def make_refine(cls, refine: str) -> str:
+        if ConceptBlueprint.is_native_concept_string_or_concept_code(concept_string_or_concept_code=refine):
+            for native_concept in NativeConceptEnum:
+                if native_concept.value == refine:
+                    return f"{SpecialDomain.NATIVE.value}.{refine}"
```
lchoquel: this loop serves no purpose, all that counts here is the test `is_native_concept_string_or_concept_code `

pipelex/core/concepts/concept_factory.py
```diff
@@ -1,168 +1,192 @@
...
-                    # Get the generated class and register it
-                    generated_class = exec_globals[code]
-                    get_class_registry().register_class(generated_class)
+                    # Register the generated class
+                    KajsonManager.get_class_registry().register_class(exec_globals[concept_code])
+
+                    # The structure_class_name of the concept is the concept_code
+                    structure_class_name = concept_code
 
                 except Exception as exc:
```
lchoquel, line 165: should be improved with non-general exception, and apply the `try` only to the lines than can raise

pipelex/core/concepts/concept_factory.py
```diff
@@ -1,168 +1,192 @@
...
+            domain=cls.make_domain_and_concept_code_from_concept_string_or_concept_code(
+                domain=domain,
+                concept_string_or_concept_code=concept_code,
+                concept_codes_from_the_same_domain=concept_codes_from_the_same_domain,
+            )[0],
+            code=cls.make_domain_and_concept_code_from_concept_string_or_concept_code(
+                domain=domain,
+                concept_string_or_concept_code=concept_code,
+                concept_codes_from_the_same_domain=concept_codes_from_the_same_domain,
+            )[1],
```
lchoquel: will look much better if you refactor `make_domain_and_concept_code_from_concept_string_or_concept_code` as requested above

lchoquel (changes requested):
Reviewed down to ConceptFactory

pipelex/cogt/content_generation/jinja2_generate.py
lchoquel: wait, I see it breaks `make ti` if we don't have that; but the bug must be fixed elsewhere so OK to keep the code to expedite the PR, but we add a TODO?

pipelex/core/concepts/concept_native.py
```diff
@@ -52,35 +22,37 @@ class NativeConcept(Enum):
...
-                return NativeConceptClass.NUMBER
-            case NativeConcept.LLM_PROMPT:
-                return NativeConceptClass.LLM_PROMPT
-            case NativeConcept.DYNAMIC:
-                return NativeConceptClass.DYNAMIC
-            case NativeConcept.PAGE:
-                return NativeConceptClass.PAGE
-            case NativeConcept.ANYTHING:
-                raise RuntimeError("NativeConcept.ANYTHING cannot be used as a content class name")
+NATIVE_CONCEPTS_DATA: Dict[NativeConceptEnum, NativeConceptEnumData] = {
```
lchoquel, line 29: don't do this with a constant dict (which can theoretically raise a KeyError at runtime when you use it like `NATIVE_CONCEPTS_DATA[NativeConceptEnum.TEXT]) and anyway it's ugly)`
instead, use match/case

pipelex/core/interpreter.py
```diff
@@ -40,6 +56,29 @@ def validate_file_path(self) -> Self:
...
+                        file.write(cleaned_content)
+                    return cleaned_content
+
+                return file_content
+
+            except Exception as exc:
+                raise ValueError(f"Failed to read TOML file '{self.file_path}': {exc}") from exc
+        else:
+            if self.file_content is None:
+                raise ValueError("file_content must be provided if file_path is not provided")
```
lchoquel: use `elif`

pipelex/core/interpreter.py
```diff
@@ -40,6 +56,29 @@ def validate_file_path(self) -> Self:
             validate_toml_content(content=self.file_content, file_path=str(self.file_path))
         return self
 
+    def get_file_content(self) -> str:
```
lchoquel, line 60: raise proper Errors, not `ValueError`

pipelex/core/concepts/concept.py
thomashebrard: The refines can now only refine Native concepts. And only 1 concept. So why the problem?
This function is now tested, with multiple different cases.
The function we had before was not tested, and was not working well.

pipelex/core/concepts/concept.py
thomashebrard: it was not used

thomashebrard: old code

pipelex/core/memory/working_memory_factory.py
```diff
@@ -6,40 +6,45 @@
...
+from pipelex.hub import get_required_concept
 
 
 class WorkingMemoryFactory(BaseModel):
     @classmethod
     def make_from_text(
         cls,
         text: str,
-        concept_str: str = NativeConcept.TEXT.code,
+        concept_string: str = SpecialDomain.NATIVE.value + "." + NativeConceptEnum.TEXT.value,
```
lchoquel, line 25: make these `concept_string` using a method of `NativeConceptEnum` like we had before with `NativeConcept.TEXT.code`

pipelex/core/pipes/pipe_blueprint.py
```diff
@@ -36,13 +32,24 @@ class PipeBlueprint(BaseModel):
 
     type: Any  # TODO: Find a better way to handle this.
     definition: Optional[str] = None
-    inputs: Optional[Dict[str, InputRequirementBlueprint]] = None
-    output: str
+    inputs: Optional[Dict[str, Union[str, InputRequirementBlueprint]]] = None
+    output: str  # concept_string_or_concept_code
```
lchoquel: instead of a comment you can leverage pydantic with an alias so that the blueprint in the TOML will be named `output` but the attribute in the BaseModel will be named `concept_string_or_concept_code `

pipelex/core/pipes/pipe_blueprint.py
```diff
@@ -36,13 +32,24 @@ class PipeBlueprint(BaseModel):
...
-    def validate_pipe_type(self) -> Self:
+    @field_validator("type", mode="after")
+    def validate_pipe_type(cls, value: Any) -> Any:
         """Validate that the pipe type is one of the allowed values."""
         allowed_types = [_type.value for _type in AllowedPipeTypes]
-        if self.type not in allowed_types:
-            raise PipeBlueprintError(f"Invalid pipe type '{self.type}'. Must be one of: {sorted(allowed_types)}")
-        return self
+        if value not in allowed_types:
+            raise PipeBlueprintError(f"Invalid pipe type '{value}'. Must be one of: {sorted(allowed_types)}")
```
lchoquel: don't sort, let's keep them in the order you have in the enum, i.e. separating Operators and Controllers

pipelex/core/concepts/concept_factory.py
thomashebrard: I don't understand

pipelex/core/concepts/concept.py
thomashebrard: Its like a subclass.
it has just field in addition

thomashebrard: This is why i introduced the arg "strict". 
Strict means that it needs to have the same fields.

pipelex/core/stuffs/stuff_factory.py
```diff
@@ -26,32 +26,52 @@ class StuffFactoryError(PipelexError):
...
+    @classmethod
+    def make_from_str(cls, str_value: str, name: str) -> Stuff:
+        return cls.make_stuff(
+            concept=ConceptFactory.make_native_concept(native_concept_data=NATIVE_CONCEPTS_DATA[NativeConceptEnum.TEXT]),
+            content=TextContent(text=str_value),
+            name=name,
+        )
+
+    @classmethod
+    def make_from_concept_string(cls, concept_string: str, name: str, content: Optional[StuffContent] = None) -> Stuff:
```
lchoquel: make `content: StuffContent` (non Optional) and avoid useless tests

pipelex/create/structured_output_generator.py
lchoquel, line 1: The methods that return False for a failed validation should just raise (proper errors).
(to be improved but non blocking)

pipelex/libraries/library_manager.py
```diff
@@ -156,20 +158,70 @@ def _get_pipelex_toml_files_from_dirs(self, dirs: List[Path]) -> List[Path]:
...
+        concepts = self._load_concepts_from_blueprint(blueprint)
+        self.concept_library.add_concepts(concepts=concepts)
+
+        # Create and load pipes
+        pipes = self._load_pipes_from_blueprint(blueprint)
+        self.pipe_library.add_pipes(pipes=pipes)
+
+        return pipes
+
+    def _load_domain_from_blueprint(self, blueprint: PipelexBundleBlueprint) -> Domain:
```
lchoquel, line 187: replace by call to DomainFactory's method

pipelex/pipe_controllers/condition/pipe_condition.py
```diff
@@ -199,16 +229,30 @@ async def _run_controller_pipe(
...
+
+        pipe_jinja2_blueprint = PipeJinja2Blueprint(
+            definition="Jinja2 template for pipe condition evaluation",
             jinja2=self.applied_expression_template,
-            inputs=self.inputs,
+            inputs=inputs_blueprint,
+            output=self.output.code,
+        )
+
+        pipe_jinja2 = PipeJinja2Factory.make_from_blueprint(
```
lchoquel, line 270: add `# TODO: use jinja2 directly without going though a pipe`

pipelex/pipe_operators/img_gen/pipe_img_gen.py
```diff
@@ -80,14 +81,31 @@ def validate_inputs(self) -> Self:
         self._validate_inputs()
         return self
 
+    @override
+    def validate_output(self):
+        if not get_concept_provider().is_compatible(
+            tested_concept=self.output,
+            wanted_concept=get_concept_provider().get_native_concept(native_concept=NativeConceptEnum.IMAGE),
```
lchoquel, line 88: we need a straight way to get native concept object, straight from NativeConceptEnum method

pipelex/pipe_works/pipe_dry.py
```diff
@@ -75,7 +69,8 @@ async def dry_run_pipes(pipes: List[PipeAbstract]) -> Dict[str, str]:
 
     start_time = time.time()
     results: Dict[str, str] = {}
-
+    pipe_library = get_pipe_provider()
```
lchoquel: should be named `pipe_provider`

pipelex/tools/misc/string_utils.py
```diff
@@ -196,3 +196,11 @@ def snake_to_capitalize_first_letter(snake_str: str) -> str:
     components = snake_str.split("_")
     phrase = " ".join(components)
     return phrase.capitalize()
+
+
+def is_snake_case(word: str) -> bool:
```
lchoquel, line 201: needs UT

pipelex/tools/misc/string_utils.py
```diff
@@ -196,3 +196,11 @@ def snake_to_capitalize_first_letter(snake_str: str) -> str:
...
+    return re.match(r"^[a-z][a-z0-9_]*$", word) is not None
+
+
+def is_pascal_case(word: str) -> bool:
```
lchoquel, line 205: needs UT

pipelex/tools/misc/toml_utils.py
lchoquel, line 1: to be redesigned with existing features of packages toml, tomlkit and taplo

tests/data/test_migrate_v0_1_0_to_v0_2_0.toml
```diff
@@ -13,14 +13,15 @@ refines = "Text"
...
+refines = "Text"
 
 [concept.AlreadyMigrated]
 definition = "This one is already migrated"
 refines = "Text"
 
 [pipe.test_pipe]
-PipeLLM = "A test pipe that generates simple text"
+type = "PipeLLM"
+description = "A test pipe that generates simple text"
```
lchoquel: I think the old syntax here was on purpose (to test migration)

lchoquel (dismissed):
some requests and some non-blocking remarks

pipelex/core/pipes/pipe_blueprint.py
thomashebrard: Good point!

pipelex/tools/misc/toml_utils.py
thomashebrard: yes

pipelex/core/concepts/concept_native.py
thomashebrard: next PR

pipelex/pipe_operators/img_gen/pipe_img_gen.py
thomashebrard: Next pr

pipelex/create/structured_output_generator.py
thomashebrard: TO be improved next PR

PR #333, opened 2025-10-03
Fix/pipe llm inputs

thomashebrard:
https://www.loom.com/share/331b03e5ab474ecca84a87d8df77b67b


pipelex/libraries/pipelines/builder/builder.plx
```diff
@@ -150,8 +150,11 @@ system_prompt = """
 You are an expert at data extraction and json formatting.
 """
 prompt_template = """
-Your job is to extract a list of ConceptSpec from these concept drafts:
+You are on a big journey to construct a pipeline, and this is one of the steps. 
+Here is the overalle mission of the user:
```
lchoquel, line 154: typo

lchoquel: But you have a point with this addition.
I just realized giving a brief (like we've done until now) inherently lacks context: if the brief says "analyze that bla for foobar", we dont want **our** prompt to ask the LLM to analyze blah for foobar. We want the LLM to design the method that will enable to analyze blah for foobar.
We are missing a clear narrative level that we should have at every stage and express in a "standard" way, like explained everywhere the same so that we can optimize it.
**We should produce this as a meta_brief right at the beginning.**
And if the brief is a mega_prompt or a codebase we would do it differently.

pipelex/pipe_operators/llm/pipe_llm_factory.py
```diff
@@ -91,8 +93,18 @@ def make_from_blueprint(
                     ),
                 )
 
-                if get_concept_provider().is_image_concept(concept=concept):
+                if Concept.are_concept_compatible(
+                    concept_1=concept, concept_2=get_concept_provider().get_native_concept(NativeConceptEnum.IMAGE), strict=True
```
lchoquel: let's avoid these compound accessors, they make ti seem like you're reaching for something far-fetched.
use `get_native_concept()` directly from the hub
if it doesn't exist, add it pls

pipelex/pipe_operators/llm/pipe_llm_factory.py
```diff
@@ -91,8 +93,18 @@ def make_from_blueprint(
...
+                ):
                     user_images.append(stuff_name)
+                elif Concept.are_concept_compatible(
+                    concept_1=concept, concept_2=get_concept_provider().get_native_concept(NativeConceptEnum.IMAGE), strict=False
+                ):
+                    # Get image field paths relative to the concept
+                    image_field_paths = get_concept_provider().find_image_field_paths(concept=concept)
```
lchoquel, line 100: seeing this call, I don't think `find_image_field_paths ` should be a method of the concept provider
you already got the `concept ` from teh provider, so maybe just make it a method of the Concept, no?

lchoquel (commented):
looks promising

pipelex/pipe_operators/llm/pipe_llm_factory.py
thomashebrard: This was my original thought.
However, this function needs to have access to other concepts to find the image path. 
Therefore needs to access the hub -> Not allowed in the class Concept

pipelex/libraries/pipelines/builder/builder.plx
thomashebrard: agree. Lets push that to the next PR

pipelex/pipe_operators/llm/pipe_llm_factory.py
thomashebrard: added in the hub

pipelex/pipe_operators/llm/pipe_llm_factory.py
lchoquel: ok, I'm not surprised

lchoquel (approved):
OK

PR #346, opened 2025-10-08
Refactor/model and template choices

lchoquel:
### 📝 Description

- rename "ocr" to "extract"
- rename all "model" attributes in blueprints/settings to simply "model"


thomashebrard (approved):
I like this "model" thing 🤩

PR #358, opened 2025-10-14
Fix/access to builder classes and pipe funcs

lchoquel:
- Fix using Pipe Builder from outside Pipelex


.cursor/rules/pytest_standards.mdc
```diff
@@ -64,11 +64,12 @@ class TestFooBar:
         # Test implementation
 ```
 
-Sometimes it can be convenient to access the test's name in its body, for instance to include into a job_id. To achieve that, add the argument `request: FixtureRequest` into the signature and then you can get the test name using `cast(str, request.node.originalname),  # type: ignore`. 
+- Never more than 1 class per test file.
+- Sometimes it can be convenient to access the test's name in its body, for instance to include into a job_id. To achieve that, add the argument `request: FixtureRequest` into the signature and then you can get the test name using `cast(str, request.node.originalname),  # type: ignore`.
```
thomashebrard, line 68: When I test 1 method, i like to have only 1 test funciton, but with different test cases in parameters

pipelex/libraries/library_manager.py
```diff
@@ -379,16 +282,25 @@ def load_libraries(
             # Only import files that contain @pipe_func decorated functions (uses AST pre-check)
             FuncRegistryUtils.register_funcs_in_folder(
                 folder_path=str(library_dir),
-                decorator_names=[pipe_func.__name__],
-                require_decorator=True,
             )
 
         # Import from pipelex package
         # Always directly import critical builder modules first (works in all installation modes)
+        log.info("About to import pipelex.builder modules for @pipe_func registration")
```
thomashebrard, line 289: debug?dev?
Why see this as a pipelex user

thomashebrard (approved):

.cursor/rules/pytest_standards.mdc
lchoquel: incorporated downstream

pipelex/libraries/library_manager.py
lchoquel: verbose downstream

PR #367, opened 2025-10-16
Feature/polish

lchoquel:
- various polish
- various refactor
- Added `pipelex run` to CLI


pipelex/pipe_operators/llm/pipe_llm.py
```diff
@@ -289,42 +279,10 @@ async def _run_operator_pipe(
...
-                        )
-                        system_prompt = self.system_prompt_to_structure or domain.system_prompt
-                        llm_prompt_2_proto = LLMPrompt(
-                            system_text=system_prompt,
-                            user_text=prompt_template_to_structure,
-                        )
-                        llm_prompt_2_factory = LLMPromptTemplate(
-                            proto_prompt=llm_prompt_2_proto,
-                        )
+                        llm_prompt_2_factory = LLMPromptTemplate.make_for_structuring_from_preliminary_text()
```
thomashebrard, line 282: if `make_for_structuring_from_preliminary_text`, why not rename `llm_prompt_for_structuring_factory`?

pipelex/cli/commands/build_cmd.py
```diff
@@ -18,6 +18,7 @@
 
 """
 Today's example:
+pipelex build pipe "Imagine a cute animal mascot for a startup based on its elevator pitch"
```
thomashebrard, line 21: I don't see how thats a workflow

pipelex/builder/bundle_header_spec.py
```diff
@@ -0,0 +1,10 @@
+from pydantic import Field
+
+from pipelex.core.stuffs.structured_content import StructuredContent
+
+
+class BundleHeaderSpec(StructuredContent):
+    domain: str = Field(description="Name of the domain of the knowledge work.")
+    description: str = Field(description="Definition of the domain of the knowledge work.")
+    system_prompt: str | None = Field(description="System prompt for the domain.")
+    main_pipe: str = Field(description="The main pipe of the domain.")
```
thomashebrard, line 10: This is equivalent to the DomainBlueprint. Why not keep our convention:
Blueprint for pipelex, Spec for pipe builder.
Therefore, DomainSpec ?

pipelex/pipe_operators/llm/pipe_llm.py
lchoquel: I applied the naming convention `make` for factory methods, and as LLMPromptTemplate is itself a Factory class, I didn't have to write LLMPrompt again

pipelex/cli/commands/build_cmd.py
lchoquel: try it, it works every time

pipelex/builder/bundle_header_spec.py
lchoquel: If did initially change its name to `DomainSpec`. But then I added `main_pipe` which is related to the bundle, not to the domain.
It smells of ambiguity for this header but it's OK to keep it like that for now.

pipelex/pipe_operators/llm/pipe_llm.py
thomashebrard: My question is more about this "2" that doesn't really mean anything.

thomashebrard (approved):

PR #375, opened 2025-10-20
Fix/inputs-api_client-implicit_memory

thomashebrard:
### 🔄 Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🧹 Code refactor
- [ ] ✅ Test update

### 📋 Checklist


pipelex/client/pipeline_response_factory.py
```diff
@@ -30,21 +56,22 @@ def make_from_pipe_output(
...
             message: Optional message providing additional information
             error: Optional error message
-
+            pipe_structures: Structure of the pipeline to execute
         Returns:
             PipelineResponse with the pipe output serialized to reduced format
 
         """
-        compact_output = None
-        if pipe_output is not None:
```
lchoquel, line 39: can't be None anymore

pipelex/client/protocol.py
```diff
@@ -30,13 +53,13 @@ class ApiResponse(BaseModel):
...
     Attributes:
-        status (str | None): Status of the API call ("success", "error", etc.)
+        status (str): Application-level status ("success", "error")
         message (str | None): Optional message providing additional information
         error (str | None): Optional error message when status is not "success"
 
     """
 
-    status: str | None = None
+    status: str = "success"
```
lchoquel: It feels strange to be "success" by default
you could make it `str | None` but without the default `= None`
this way, a status value must be pased when creating the  instance

pipelex/core/concepts/structure_generator.py
```diff
@@ -188,23 +188,20 @@ def _get_python_type_from_blueprint(self, field_blueprint: ConceptStructureBluep
                 # Recursively handle item types if they're FieldType enums
                 try:
                     item_type_enum = ConceptStructureBlueprintFieldType(item_type)
-                    # Create a temporary blueprint for the item type
-                    temp_blueprint = ConceptStructureBlueprint(description="temp", type=item_type_enum)
-                    item_type = self._get_python_type_from_blueprint(temp_blueprint)
+                    if item_type_enum == ConceptStructureBlueprintFieldType.DICT:
+                        temp_blueprint = ConceptStructureBlueprint(description="temp", type=item_type_enum, key_type="str", value_type="Any")
```
lchoquel: "temp" is very lose here
Why not name more specifically `item_blueprint`?

lchoquel (dismissed):
small remarks

pipelex/client/pipeline_response_factory.py
thomashebrard: Yes if It does if the pipeline fails. This way we can still return information like the workflow id, the datetimes and so on

lchoquel: My comment doesn't make sense

PR #399, opened 2025-10-27
Docs/rules to call cli

lchoquel:
- Agent rules to call direct CLI instead of make
- Avoid failure in sub-pipe when there is not main stuff from subpipe
- Cleanup smelly code

### DIFF
--- a/.cursor/rules/run_pipelex.mdc
+++ b/.cursor/rules/run_pipelex.mdc
@@ -6,6 +6,21 @@ globs:
 ---
 # Guide to execute a pipeline and write example code
 
+## Prerequisites: Virtual Environment
+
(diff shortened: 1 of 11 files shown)


.cursor/rules/run_pipelex.mdc
```diff
@@ -6,6 +6,21 @@ globs:
 ---
 # Guide to execute a pipeline and write example code
 
+## Prerequisites: Virtual Environment
+
+**CRITICAL**: Before running any `pipelex` commands or `pytest`, you MUST activate the appropriate Python virtual environment. Without proper venv activation, these commands will not work.
```
thomashebrard, line 11: Isn't he going to do it EVERY time?

thomashebrard (approved):

.cursor/rules/run_pipelex.mdc
lchoquel: that is what I hope

PR #400, opened 2025-10-27
Fix/cli init and bugs with pipe controllers

lchoquel:
- cli init
- fix bugs with pipe controllers and subpipe

### DIFF
--- a/pipelex/pipe_controllers/pipe_controller.py
+++ b/pipelex/pipe_controllers/pipe_controller.py
@@ -7,6 +7,7 @@ from pipelex import log
 from pipelex.core.memory.working_memory import WorkingMemory
 from pipelex.core.pipes.pipe_abstract import PipeAbstract
 from pipelex.core.pipes.pipe_output import PipeOutput
+from pipelex.exceptions import WorkingMemoryStuffNotFoundError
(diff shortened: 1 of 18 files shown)


thomashebrard (dismissed):

pipelex/pipe_controllers/pipe_controller.py
```diff
@@ -36,6 +50,9 @@ async def run_pipe(
         )
         job_metadata.update(updated_metadata=updated_metadata)
 
+        # check we have the required inputs in the working memory
+        self._validate_inputs_in_memory(working_memory=working_memory)
+
```
thomashebrard, line 55: This should be done by the dry run, not at run time

lchoquel: We do it for dry run already, which detects internal errors, but at run time, there is a new possibility to have missing inputs: the user executes the pipeline with missing stuff in the provided inputs.
This new validation will make sure we detect it right away, thanks to the main Pipe (usually a Sequence) instead of crashing when the variable is missing for a prompt template or other ugly error.

thomashebrard (approved):

PR #419, opened 2025-11-01
Fix a Pretty print and httpx bugs

### LINKED ISSUE #421: Unable to get the pipeline working with ollama model
### Before submitting

- [x] I'm using the **latest released** version of the library
- [x] I've searched [open issues](issues?q=is%3Aissue%20state%3Aopen%20type%3ABug) and found no duplicate

### Describe the bug, tell us what went wrong

With the official ollama.toml provided https://github.com/Pipelex/pipelex/blob/main/.pipelex/inference/backends/ollama.toml, the execution errors out with the following message:
`OpenAIError: The api_key client option must be set either by passing api_key to the client or by setting the OPENAI_API_KEY environment variable`

### Reproduction snippet

uv run pipelex run pipe

### Library version

v0.14.3

### Python version

3.13.2

### Operating system

_No response_

### Stack trace / error output

```shell

```

### Additional context & screenshots

_No response_

### Would you like to help fix this issue?

Yes, I'd like to contribute

lchoquel:
- Pretty print content bug
- add httpx user agent so it doesn't get rejected by sites like Wikipedia
- Fix #421

### DIFF
--- a/pipelex/plugins/openai/openai_factory.py
+++ b/pipelex/plugins/openai/openai_factory.py
@@ -48,7 +48,7 @@ class OpenAIFactory:
             msg = f"Plugin '{plugin}' is not supported by OpenAIFactory"
             raise OpenAIFactoryError(msg) from exc
 
-        # We have a workaround here
+        # We have a workaround here:
         # OpenAI can be used without any API key (for instance when pointing to local Ollama) but the SDK,
         # as it is, raises if there is not API key (api_key is None and there is not env var).
         # But it works fine with an empty string.
--- a/pipelex/plugins/openai/openai_factory.py
+++ b/pipelex/plugins/openai/openai_factory.py
@@ -48,6 +48,11 @@ class OpenAIFactory:
             msg = f"Plugin '{plugin}' is not supported by OpenAIFactory"
             raise OpenAIFactoryError(msg) from exc
 
+        # We have aworkaround here: OpenAI's SDK can be used without any API key (for instance when pointing to local Ollama) but the SDK,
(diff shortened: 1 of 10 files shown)


pipelex/plugins/openai/openai_factory.py
```diff
@@ -64,13 +64,13 @@ def make_openai_client(
...
-                    api_key=backend.api_key,
+                    api_key=backend.api_key or "",
                     base_url=backend.endpoint,
                 )
             case OpenAISdkVariant.OPENAI_ALT_IMG_GEN:
                 log.verbose(f"Making AsyncOpenAI client with endpoint: {backend.endpoint}")
                 the_client = openai.AsyncOpenAI(
-                    api_key=backend.api_key,
+                    api_key=backend.api_key or "",
                     base_url=backend.endpoint,
```
thomashebrard, line 79: Sending an empty string as an api key should not be possible, it should raise before.
I understand this is the bug you were talking about earlier. But the it should be properly fixed

thomashebrard (commented):
Some comments

pipelex/plugins/openai/openai_factory.py
lchoquel: The bug is that OpenAI's SDK **can** be used without any key (for instance when pointing to local Ollama) but the SDK, as it is, raises if there is not key (api_key is None and there is not env var). So in a sense the bug is inside the OpenAI SDK.
On our end, it makes perfect sense to have this `or ""` workaround in the OpenAIFactory code
I'll add a comment to clarify.

thomashebrard: got it

PR #431, opened 2025-11-06
Feature/merge model fallback

lchoquel:
- merge


pipelex/cogt/img_gen/img_gen_setting.py
```diff
@@ -20,7 +19,7 @@ class ImgGenSetting(ConfigModel):
     def validate_quality_or_nb_steps(self) -> Self:
         if self.quality is not None and self.nb_steps is not None:
             msg = "ImgGenSetting cannot have both 'quality' and 'nb_steps' specified. Use one or the other."
-            raise ImgGenSettingsValidationError(msg)
+            raise ValueError(msg)
```
thomashebrard, line 22: you can still raise ImgGenSettingsValidationError if its a subclass of ValueError

thomashebrard (approved):

pipelex/cogt/img_gen/img_gen_setting.py
lchoquel: I know, I removed it because it caused a cyclic import and was't worth the fight at that point

PR #454, opened 2025-11-15
Refactor/use groq without specific llm worker

lchoquel:
- Improve setting for structure method
- Thanks to that, Groq now uses the openai LLMWorker
- I also moved the unit tests back to tests/unit/pipelex instead of tests/unit
- And added a tool to diff between directories + Makefile target to check diffs between `.pipelex` and `pipelex/kit/configs` -> meant to be added to CI checks too


pipelex/plugins/google/google_llm_worker.py
```diff
@@ -56,6 +53,28 @@ def __init__(
...
+        except RuntimeError:
+            # No running event loop, we can safely use asyncio.run()
+            try:
+                asyncio.run(self.genai_async_client.aclose())
+            except Exception as exc:
+                # Log but don't fail teardown if cleanup has issues
+                log.debug(f"Error closing Google async client during teardown: {exc}")
+        except Exception as exc:
+            # Log but don't fail teardown if cleanup has issues
+            log.debug(f"Error during Google async client teardown: {exc}")
```
thomashebrard, line 95: Isn't this going to close other async loops if existing ? that is not related to the worker?

thomashebrard (dismissed):

pipelex/plugins/google/google_llm_worker.py
lchoquel: No it only closed the genai client but the code was bad practice nonetheless, so I replaced it with best practice

PR #463, opened 2025-11-21
Fix/remarks

thomashebrard:
### 🔄 Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🧹 Code refactor
- [ ] ✅ Test update

### 📋 Checklist


pipelex/builder/flow_factory.py
```diff
@@ -53,7 +53,7 @@ def make_from_bundle_blueprint(bundle_blueprint: PipelexBundleBlueprint) -> Flow
 
         if bundle_blueprint.pipe:
             for pipe_code, pipe_blueprint in bundle_blueprint.pipe.items():
-                if pipe_blueprint.pipe_category == AllowedPipeCategories.PIPE_CONTROLLER:
+                if pipe_blueprint.pipe_category == PipeCategory.PIPE_CONTROLLER:
```
lchoquel: use `PipeCategory.is_controller_by_str`

pipelex/core/pipes/pipe_blueprint.py
```diff
@@ -120,22 +120,22 @@ def ordered_pipe_dependencies(self) -> list[str] | None:
...
+        if value not in PipeCategory.value_list():
+            msg = f"Invalid pipe category '{value}'. Must be one of: {PipeCategory.value_list()}"
             raise ValueError(msg)
         return value
 
     @model_validator(mode="after")
     def validate_pipe_category_based_on_type(self) -> Self:
-        if self.pipe_category != AllowedPipeTypes(self.type).category:
+        if self.pipe_category != PipeType(self.type).category:
             msg = f"Invalid pipe category '{self.pipe_category}'. Must be one of: {self.type.category}"
```
lchoquel: wrong msg, it should be about the inconsistency

pipelex/core/pipes/pipe_blueprint.py
```diff
@@ -154,39 +154,36 @@ def validate_output(self):
 
     @final
     def generic_validate_inputs(self):
-        if self.inputs is None:
```
lchoquel, line 157: it was better to exit early when there are no inputs to validate, juste the best way to do it is:
```
if not self.inputs:
    return
```

pipelex/core/pipes/pipe_blueprint.py
```diff
@@ -154,39 +154,36 @@ def validate_output(self):
...
-            return
-
         # Pattern allows: ConceptName, domain.ConceptName, ConceptName[], ConceptName[N]
         multiplicity_pattern = MUTLIPLICITY_PATTERN
```
lchoquel: I don't see the point of creating a local var from this,
you can move your comment to right above the `match = re.match(…` where it's used and it should use the constant directly

pipelex/libraries/concept/concept_library.py
```diff
@@ -39,12 +38,21 @@ def reset(self):
         self.teardown()
         self.setup()
 
+    # TODO: Rethink the make_empty of libraries. It lakes sense with the setup and teardown methods
```
lchoquel: typo: "makes"

pipelex/libraries/library.py
```diff
@@ -15,6 +16,8 @@ class Library(BaseModel):
     This represents a complete set of Pipelex definitions (domains, concepts, pipes)
     that can be loaded and used together, typically for a single pipeline run.
 
+    Limitations: It misses the Func Registry library and Class Registry library
```
lchoquel: "It lacks"

Makefile
```diff
@@ -482,11 +482,11 @@ docs-deploy: env
 ### SHORTHANDS
 ##########################################################################################
 
-c: format lint pyright mypy
+c: format lint pyright pylint mypy
```
lchoquel: no, I don't want pylint in `make c`, it's too slow,
we already decided that,
we have it in `make check`

lchoquel (changes requested):
simpel remarks

pipelex/core/pipes/pipe_blueprint.py
thomashebrard: I did that because if i return None if self.inputs is None, it will NONE run the specific pipe validation function `        self.validate_inputs()` that is below

lchoquel (approved):
Great

PR #474, opened 2025-11-24
Fix/better validation error

thomashebrard:
Cleaned Errors, Removed unused Erros


pipelex/core/interpreter/helpers.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Returns:
+        ValidationErrorScope - defaults to BUNDLE if scope cannot be determined
+    """
+    if not loc:
+        return ValidationErrorScope.BUNDLE
+
+    first = str(loc[0])
+
+    if ValidationErrorScope.is_pipe_scope(scope=first):
```
lchoquel: convert it to ValidationErrorScope and then use a proper match/case

pipelex/core/interpreter/helpers.py
```diff
@@ -0,0 +1,127 @@
...
+    Args:
+        loc: Location tuple from Pydantic validation error
+
+    Returns:
+        ValidationErrorScope - defaults to BUNDLE if scope cannot be determined
+    """
+    if not loc:
+        return ValidationErrorScope.BUNDLE
+
+    first = str(loc[0])
```
lchoquel: `first` as a var name doesn't mean much, shouldn't it be named `scope`?

pipelex/core/interpreter/validation_error_categorizer.py
```diff
@@ -0,0 +1,166 @@
...
+    """Categorize a BLUEPRINT validation error and create structured error data or return None if the error cannot be categorized.
+
+    Args:
+        blueprint_dict: The blueprint dict being validated (for context extraction)
+        error: Pydantic error from PipelexBundleBlueprint.model_validate()
+
+    Returns:
+        PipelexBundleBlueprintValidationErrorData with all relevant fields populated, or None if error cannot be categorized
+    """
+    domain = blueprint_dict.get("domain") if blueprint_dict else None
```
lchoquel: TODO: replace magic string "domain" by StrEnum,
same for "source", "loc"… all the keys we `get()`. They can all be in the same StrEnum I guess, named something like `ErrorCatKey` (to keep it short)

pipelex/core/interpreter/validation_error_categorizer.py
```diff
@@ -0,0 +1,166 @@
...
+    last_step_output_concept: str | None = None
+    expected_output_concept: str | None = None
+
+    if "last step" in message:
+        # Extract: "of the last step 'my_pipe'"
+        try:
+            if "last step '" in message:
+                start = message.index("last step '") + len("last step '")
+                end = message.index("'", start)
+                last_step_pipe_code = message[start:end]
```
lchoquel: TODO: code a helper for this value extraction technique, used 3 times here (DRY rule)

pipelex/pipe_controllers/parallel/pipe_parallel.py
```diff
@@ -230,22 +224,17 @@ async def _dry_run_controller_pipe(
...
-            msg = f"Dry run failed: missing required inputs: {missing_input_names}"
-            log.error(f"Dry run failed: missing required inputs: {missing_input_names}")
-            raise DryRunMissingInputsError(
-                message=msg,
-                pipe_type=self.__class__.__name__,
-                pipe_code=self.code,
-                missing_inputs=missing_input_names,
-            )
+            msg = f"Dry run failed for pipe '{self.code}' (PipeParallel): missing required inputs: {missing_input_names}"
+            log.error(msg)
```
lchoquel: I doubt we need to log before raising?

pipelex/pipe_controllers/sequence/pipe_sequence.py
```diff
@@ -48,17 +48,18 @@ def validate_output_with_library(self):
         last_step_output_pipe = get_required_pipe(pipe_code=self.sequential_sub_pipes[-1].pipe_code)
```
lchoquel: it would read better here to assign var with `self.sequential_sub_pipes[-1].pipe_code` (and use it twice)

btw should we also check sequential_sub_pipes is not empty?

pipelex/pipe_operators/func/pipe_func.py
```diff
@@ -137,23 +136,21 @@ async def _dry_run_operator_pipe(
...
-            msg = f"Dry run failed for PipeFunc (function '{self.function_name}'): missing required inputs: {missing_input_names}"
-            log.error(f"Dry run failed: missing required inputs: {missing_input_names}")
-            raise DryRunMissingInputsError(
-                message=msg,
-                pipe_type=self.__class__.__name__,
-                pipe_code=self.code,
-                missing_inputs=missing_input_names,
-            )
+            msg = f"Dry run failed for pipe '{self.code}' (PipeFunc): missing required inputs: {missing_input_names}"
+            log.error(msg)
```
lchoquel: useful to log error?

pipelex/pipeline/validate_bundle.py
```diff
@@ -1,118 +1,67 @@
...
+        self.pipe_concept_instantiation_errors = pipe_concept_instantiation_errors or []
+
+        # Dry run errors
         self.dry_run_error_message = dry_run_error_message
+
         super().__init__(message)
 
+    @property
+    def pipe_validation_error_data(self) -> list[PipesAndConceptValidationErrorData]:
+        """Backwards compatibility: combine pipe validation and instantiation errors.
```
lchoquel, line 59: Backwards compatibility => TODO refactor so we don't need this anymore?

lchoquel (changes requested):
Some remarks

pipelex/pipe_controllers/sequence/pipe_sequence.py
thomashebrard: done

pipelex/core/pipes/handle_pipe_errors.py
```diff
@@ -124,9 +99,9 @@ def _handle_pipe_errors(
         PipesAndConceptValidationErrorData with all context populated
     """
     # Extract data from error
-    loc = error.get("loc", ())
-    message = error.get("msg", "Unknown validation error")
-    pydantic_type = error.get("type", "")
+    loc = error.get(ErrorCatKey.LOC.value, ())
```
lchoquel, line 102: you don't need the ".value" part with StrEnum, which is equivalent to `str`

lchoquel (approved):
minor non-blocking remark

PR #481, opened 2025-11-26
fix/remove-pyproject.toml-from-build

thomashebrard:
### Changed

- Modified the GHA `version-check.yml` so that the check of the version is only applying to release branches.
- Removed the `pyproject.toml` file from the build.


lchoquel (approved):
OK

PR #482, opened 2025-11-26
fix/deep_update-overrides-arrays

thomashebrard:
### Unreleased

**Highlights:** - Previously, in the pipelex config `.pipelex/pipelex.toml`, when an array was overridden, the new array was concatenated to the old array. Now, the new array overrides the old array.

### Refactored

- The `find_files_in_dir` function was coded in 3 different places, now it's in `pipelex/tools/misc/file_utils.py`, and accepts `excluded_dirs`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,16 @@
 
 ### Unreleased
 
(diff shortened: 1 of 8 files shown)


lchoquel (approved):
LGTM

PR #483, opened 2025-11-26
Fix/remove implicit concepts

thomashebrard:
- No more implicit concepts. A concept reference has a domain and a code. If there is no domain, it should be a native concept, or it is declared in the same bundle.
- Refactored the Pipe factories: Centralized everything in the `PipeFactory` class.


pipelex/builder/builder.plx
```diff
@@ -305,7 +305,7 @@ prompt = """
 [pipe.write_bundle_header]
 type = "PipeLLM"
 description = "Write the bundle header."
-inputs = { brief = "UserBrief", pipe_signatures = "PipeSignature" }
+inputs = { brief = "UserBrief", pipe_signatures = "pipe_design.PipeSignature" }
```
lchoquel: could you put the pipe_design.plx and concept.plx also on the same "builder" domain?

pipelex/core/pipes/pipe_factory.py
```diff
@@ -48,10 +108,46 @@ def make_from_blueprint(
...
+            inputs=parsed_inputs,
+            output=parsed_output,
             blueprint=blueprint,
-            concept_codes_from_the_same_domain=concept_codes_from_the_same_domain,
         )
-        return pipe_from_blueprint
+        return pipe
+
+    @classmethod
+    def _parse_output(
```
lchoquel: Misleading: it sounds like you're parsing the output of a pipe…
you could rename to `_parse_output_concept_string()`

lchoquel (changes requested):
Small remarks + could you keep the name `pipe_code` everywhere we have a pipe's code

lchoquel (dismissed):
OK with ideally pipe_design pipes in pipe_design.plx

lchoquel (approved):
great

PR #491, opened 2025-11-27
Feature/ruff preview

thomashebrard:


tests/integration/pipelex/pipes/conftest.py
```diff
@@ -1,15 +1,15 @@
...
+import pytest
 
 from pipelex.core.pipes.pipe_output import PipeOutput
 from pipelex.tools.misc.json_utils import save_as_json_to_path
 
 
-@pytest_asyncio.fixture  # pyright: ignore[reportUntypedFunctionDecorator, reportUnknownMemberType]
-async def save_working_memory() -> Callable[[PipeOutput, str], Awaitable[None]]:
-    async def _save_working_memory(pipe_output: PipeOutput, result_dir_path: str) -> None:
+@pytest.fixture  # pyright: ignore[reportUntypedFunctionDecorator, reportUnknownMemberType]
```
lchoquel: I think you can drop this `# pyright: ignore` if not async

pipelex/builder/pipe/pipe_spec.py
```diff
@@ -125,7 +125,7 @@ def validate_pipe_code_syntax(cls, pipe_code: str) -> str:
     def to_blueprint(self) -> PipeBlueprint:
         return PipeBlueprint(
             description=self.description,
-            inputs=self.inputs if self.inputs else None,
+            inputs=self.inputs or None,
```
lchoquel, line 128: `inputs=self.inputs,` is enough for the same effect

pipelex/pipeline/track/flow_chart.py
```diff
@@ -197,8 +197,12 @@ def generate_subgraph_lines(self, graph_tree: GraphTree) -> list[str]:
...
                 subgraph_lines.extend(_indent_line(mermaid_line, 2) for mermaid_line in node_lines)
             else:
-                subgraph_lines.append(_indent_line(f'subgraph "{subgraph_name}"', 1))
-                subgraph_lines.append(_indent_line("direction LR", 1))
+                subgraph_lines.extend(
+                    [
+                        _indent_line(f'subgraph "{subgraph_name}"', 1),
+                        _indent_line("direction LR", 1),
+                    ]
+                )
```
lchoquel: no, I think it it was better before, the indentation inside the list makes it less readable

pipelex/plugins/mistral/mistral_factory.py
```diff
@@ -147,7 +147,7 @@ def make_extracted_image_from_page_from_mistral_ocr_image_obj(
             top_left_y=mistral_ocr_image_obj.top_left_y,
             bottom_right_x=mistral_ocr_image_obj.bottom_right_x,
             bottom_right_y=mistral_ocr_image_obj.bottom_right_y,
-            base_64=mistral_ocr_image_obj.image_base64 if mistral_ocr_image_obj.image_base64 else None,
+            base_64=mistral_ocr_image_obj.image_base64 or None,
```
lchoquel, line 150: you can remove the `or None`

pipelex/tools/log/log_formatter.py
```diff
@@ -15,7 +15,7 @@ def emoji_for_channel(channel_name: str) -> str | None:
     }
 
     emoji = channel_emojis.get(channel_name)
-    if emoji == "":
+    if not emoji:
         # blank emoji is OK
```
lchoquel, line 19: no, you changed the behavior:
there is a difference between `== ""` and being falsy, and the comment was here to note it

pipelex/tools/misc/markdown_utils.py
```diff
@@ -22,8 +22,7 @@ def convert_to_markdown(data: Any, level: int = 1, is_pretty: bool = False, key:
             converted_value = convert_to_markdown(data=_value, level=level + 1, key=_key)
             converted_value_nb_lines = len(converted_value.split("\n"))
             if converted_value_nb_lines > 1:
-                dict_result_lines.append(converted_line)
-                dict_result_lines.append(converted_value)
+                dict_result_lines.extend([converted_line, converted_value])
```
lchoquel: less readable

lchoquel (changes requested):
Some important remarks
and also:
- revert the replacements of `== ""` by `not`, it's a bad move it changes the behavior in subtle ways that would require careful reviewing, we don't have time for that
- replacing 3.14 by math.pi in tests is useless, the `import math` is a bad move if we're not really doing math

pipelex/plugins/mistral/mistral_factory.py
thomashebrard:
```
Argument of type "OptionalNullable[str]" cannot be assigned to parameter "base_64" of type "str | None" in function "__init__"
  Type "OptionalNullable[str]" is not assignable to type "str | None"
    Type "Unset" is not assignable to type "str | None"
      "Unset" is not assignable to "str"
      "Unset" is not assignable to "None"basedpyright[reportArgumentType](https://docs.basedpyright.com/v1.32.1/configuration/config-files/#reportArgumentType)
```

pipelex/tools/log/log_formatter.py
```diff
@@ -41,7 +41,7 @@ class EmojiLogFormatter(logging.Formatter):
     def format(self, record: logging.LogRecord):
         log_fmt: str
         emoji = emoji_for_channel(record.name)
-        if not emoji:
+        if emoji == "":  # noqa: PLC1901
```
lchoquel: disable the rule so that we don't need a noqa here

pipelex/pipeline/track/flow_chart.py
```diff
@@ -197,12 +197,8 @@ def generate_subgraph_lines(self, graph_tree: GraphTree) -> list[str]:
             if subgraph_name == "root":
                 subgraph_lines.extend(_indent_line(mermaid_line, 2) for mermaid_line in node_lines)
             else:
-                subgraph_lines.extend(
-                    [
-                        _indent_line(f'subgraph "{subgraph_name}"', 1),
-                        _indent_line("direction LR", 1),
-                    ]
-                )
+                subgraph_lines.append(_indent_line(f'subgraph "{subgraph_name}"', 1))  # noqa: FURB113
```
lchoquel: disable the rule so that we don't need a noqa here

lchoquel (changes requested):
remarks

lchoquel (approved):
Good

PR #495, opened 2025-12-01
fix/excluded-dirs-pipe-builder

thomashebrard:
### Added

- Added `exception_dirs` parameter to the `find_files_in_dir` function. This allows specific directories to be included in the search even when they are nested within excluded directories. For example, you can now exclude `.venv` while still including `.venv/lib/python3.11/site-packages/pipelex` for loading Pipelex libraries from installed packages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+### Unreleased
+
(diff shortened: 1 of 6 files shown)


lchoquel (approved):
LGTM

PR #598, opened 2026-01-19
fix/bug-missing-inputs

thomashebrard:
Fix bug missing inputs at runtime: Add validation before and after run pipe (for live and dry run)


lchoquel (approved):
include validate_before_run in the try block

PR #608, opened 2026-01-22
fix/GHA-pyright

thomashebrard:
- The recommendation for linting dev dependencies is to pin exact versions
- make merge-check-pyright wasn't pointing to the right configuration
- fixed a bug with the latest pyright release (more strict)


lchoquel (approved):
LGTM

PR #610, opened 2026-01-22
fix/fast-cli

thomashebrard:
Moved the import of instructor to runtime and behind typecheckers


thomashebrard:
@cursor

pipelex/plugins/openai/openai_completions_llm_worker.py
```diff
@@ -2,7 +2,6 @@
 
 import instructor
```
lchoquel: `instructor` is still imported here so…

lchoquel (commented):
remark

lchoquel (approved):
LGTM

PR #630, opened 2026-02-02
Feature/tweaks moad

lchoquel:
> [!NOTE]
> **Medium Risk**
> Touches core model routing/fallback and content/storage URL semantics (`display_link`→`public_url`), which can affect inference selection and downstream integrations; changes are broad but mostly mechanical with added test coverage.
> 
> **Overview**
> **Inference/config behavior changes:** adds a first-class `internal` (software-only) backend and updates model resolution so default-routed models can fall back to `internal` even without explicit `fallback_order`; updates extract deck defaults/aliases (moving from `$extract-*` presets to `@default-*` aliases) and adjusts builder talent mappings accordingly.
> 
> **Templating & content changes:** renames `display_link` to `public_url` across `ImageContent`/`DocumentContent`, storage providers, JS helpers, and tests; improves `ImageContent` rendering output; and makes `ListContent`/`StuffArtefact` directly iterable/indexable in Jinja2 (with new unit/integration coverage).
> 
> **ImgGen/docs tweaks:** clarifies `PipeImgGen` requires an explicit `prompt` field in docs; expands Google image-gen aspect-ratio/size mapping and adds Gemini-specific extra request config; relaxes OpenAI completions img-gen square-only restriction and stops fabricating image sizes when APIs don’t return them.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e0142c4407eef7d49b03d2a0e88c7d877c60123b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>


pipelex/plugins/openai/openai_completions_img_gen_worker.py
```diff
@@ -121,7 +116,7 @@ async def _gen_image(
         return GeneratedImageRawDetails(
             actual_url=actual_url,
             base64_str=base64_str,
-            size=ImageSize(width=1024, height=1024),
+            size=None,
```
thomashebrard, line 121: Why?

pipelex/pipe_operators/img_gen/pipe_img_gen.py
```diff
@@ -193,7 +193,7 @@ async def _live_run_operator_pipe(
         the_content: StuffContent
         nb_images: int
         if isinstance(applied_output_multiplicity, bool):
-            if self.output_multiplicity:
+            if self.output_multiplicity is True:
```
thomashebrard, line 196: Good catch

thomashebrard (approved):

tests/integration/pipelex/fixtures/img_gen_fixtures.py
```diff
@@ -31,18 +31,18 @@
...
+        # ImgGenJobParams(
+        #     aspect_ratio=AspectRatio.LANDSCAPE_4_3,
+        #     background=Background.OPAQUE,
+        #     quality=Quality.HIGH,
+        #     guidance_scale=2.5,
+        #     is_moderated=None,
+        #     safety_tolerance=1,
+        #     is_raw=None,
+        #     output_format=ImageFormat.JPEG,
+        # ),
```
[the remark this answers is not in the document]
lchoquel: Fixed

pipelex/plugins/openai/openai_completions_img_gen_worker.py
lchoquel: It was hard-coded before because we didn't support the various aspect ratios in that worker. Now we do but we're in a situation where the APIs don't return the size information, so we would have to check it out on the image itself.

But why would we do that? Our internal implementation does not require that we include the size in this object, so if down the line we have some other module that needs it, it can still decide that if the size is none, then it would be its responsibility to detect the size using Pillow or something. 

I'll add a comment

pipelex/pipe_operators/img_gen/pipe_img_gen.py
lchoquel: this "thrutiness" system in python is a curse

pipelex/plugins/openai/openai_completions_img_gen_worker.py
```diff
@@ -63,9 +61,6 @@ async def _gen_image(
                 )
                 raise ImgGenParameterError(msg)
             image_format = ImageFormat.JPEG
-        if img_gen_job.job_params.aspect_ratio != AspectRatio.SQUARE:
```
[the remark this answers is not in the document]
lchoquel: Blackbox AI img gen will be removed

PR #645, opened 2026-02-05
Fix/html and pipe func

thomashebrard:
- **PipeFunc Return Type Validation**: Added validation to ensure that a `PipeFunc` function's return type matches the output concept's structure class.
 - **StructuredContent.rendered_html()**: Now recursively calls `rendered_html()` on nested `StuffContent` fields instead of using json2html conversion. Also skips `None` values and uses HTML table format.
 - Fixed structure generation failing with `PydanticUserError` when a concept structure references native concepts (e.g., `native.Html`). The generator now properly resolves native concept refs to their content classes (e.g., `HtmlContent`) with correct imports.


pipelex/cli/commands/build/structures_cmd.py
```diff
@@ -280,17 +280,37 @@ def build_structures_command(
...
     async def _build_structures_cmd():
         target_path = Path(target).resolve()
 
         if not target_path.exists():
             typer.secho(f"❌ Target does not exist: {target_path}", fg=typer.colors.RED, err=True)
             raise typer.Exit(1)
 
+        # Convert library_dir strings to Path objects
+        library_dirs_paths: list[Path] | None = [Path(lib_dir) for lib_dir in library_dir] if library_dir else None
+
```
lchoquel, line 310: shouldn't you call `resolve_library_dirs()` like we do in the other commands?

pipelex/core/concepts/structure_generation/generator.py
```diff
@@ -353,17 +353,31 @@ def _resolve_concept_ref_to_type(self, concept_ref: str | None) -> str:
...
               tracks for mocking, returns class name directly (for file generation)
             - Otherwise: returns a forward reference string (for runtime - resolved via model_rebuild)
         """
         if not concept_ref:
             return "Any"
 
+        # Handle native concepts (e.g., "native.Html" -> HtmlContent)
+        if NativeConceptCode.is_native_concept_ref_or_code(concept_ref):
+            # Extract the concept code (e.g., "Html" from "native.Html")
+            concept_code = concept_ref.split(".")[-1] if "." in concept_ref else concept_ref
```
lchoquel: this should use a helper from concept or native concept or something

lchoquel (changes requested):
small remarks

lchoquel (approved):
LGTM

PR #647, opened 2026-02-05
Fix/build structures should only validate concepts

thomashebrard:
- **`pipelex build structures` Command**: Now uses a lightweight loading mechanism that only processes domains and concepts, skipping pipe loading and validation entirely. This fixes the chicken-and-egg problem where structure generation would fail due to pipe validation errors before the structures were even created. Added `--force` / `-f` flag to regenerate all structures without checking if classes already exist.


lchoquel (approved):
LGTM

PR #707, opened 2026-02-25
fix/-L library

thomashebrard:
. Method discovery from -L directories (pipelex/cli/installed_methods.py)

  - Added _discover_method_at() helper to check a single directory for METHODS.toml
  - Added discover_methods_from_library_dirs() — recursively finds methods from -L directories:
    - If the dir itself has METHODS.toml, it's treated as a method
    - Otherwise, rglob(MANIFEST_FILENAME) finds methods at any depth below (so -L . works)
  - Updated find_method_by_name() with optional library_dirs parameter — merges library dir methods with standard methods, deduplicating by path

  2. Threading library_dirs through resolution (pipelex/cli/method_resolver.py)

  - Added library_dirs parameter to resolve_method_target(), passed to find_method_by_name()

  3. All 8 callsites updated

  - Each run method, validate method, build runner/inputs/output method, and agent CLI equivalents now pass their -L parameter to resolve_method_target()

  4. Fixed premature "Executing/Validating" messages

  - _run_core.py: Removed the premature "🚀 Executing pipe" message that printed before the pipe was verified to exist
  - _validate_core.py: Moved "Validating pipe" message to after get_required_pipe() succeeds

  5. Tests (tests/unit/pipelex/cli/test_installed_methods.py)

  - 10 tests covering: direct discovery, subdirectory scanning, deep nesting (-L . scenario), deduplication, name fallback, find_method_by_name with library dirs, not-found errors, duplicate errors

### DIFF
--- a/pipelex/cli/agent_cli/commands/inputs/method_cmd.py
+++ b/pipelex/cli/agent_cli/commands/inputs/method_cmd.py
@@ -45,6 +45,7 @@ def inputs_method_cmd(
     pipe_code, method_library_dirs, method = resolve_method_target(
         method_name=name,
         pipe_override=pipe,
+        library_dirs=library_dir,
     )
     bundle_path: Path | None = None
     if method.mthds_files:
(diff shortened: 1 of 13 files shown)


lchoquel (approved):
OK

Issue #521, opened 2025-12-15
Substitute image names in prompts for image collections


lchoquel:
### Before submitting

- [x] I'm using the **latest released** version of the library
- [x] I've searched [open issues](issues?q=is%3Aissue%20state%3Aopen%20type%3ABug) and found no duplicate

### Describe the bug, tell us what went wrong

To include an image input into an LLM prompt, you reference it in the prompt. And then the image gets sent to the LLM and the in the prompt text, we substitute the image name by a image number as in `Image-3`.
This is broken when the images come from a collection of images like a `ListContent` or a simple `list[]` nested in a structured object.
We must find a way to avoid including their names as it can be misleading for the LLM.

### Reproduction snippet

Run integration test with inference enabled:
`make tip TEST=test_analyze_image_collection`

### Library version

0.17.3

### Python version

3.11

### Stack trace / error output

```shell

```

### Would you like to help fix this issue?

Not at this time

PR #917, opened 2026-05-18
Fix/521 image collection placeholders

matthewYang08:
### 🔗 Related Issues
#521 
### 🔄 Type of Change
- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 💥 Breaking change
- [ ] 📚 Documentation update
- [ ] 🧹 Code refactor
- [ ] ✅ Test update

### 📋 Checklist

---
## Summary by cubic
Hide outer field names for image-only collections in prompt output and tighten image-field discovery across `list`, `tuple`, `dict`, and `ListContent` (including Optional/Union types). Keep dict item keys and labels for single-image fields. Addresses Linear 521.

- **Bug Fixes**
  - Rendering: omit outer names for image-only collections; keep `dict` item keys; keep field names when collections mix images and non-images.
  - Discovery: unified checks handle `Union`/`Optional`, nested generics, `dict` values only (ignore key types), and `ListContent` item types via Pydantic metadata.
  - Tests: added cases for dicts, tuples, `ListContent[ImageContent]`, and Optional items in containers; updated expected prompt output.

<sup>Written for commit f8b8fec68b922089742f3508508ffd570ffd72ca. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Pipelex/pipelex/pull/917?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


matthewYang08:
Hi maintainers, this PR is ready for CI.
Could you please approve workflows so checks can run? Thanks.

pipelex/core/stuffs/image_field_search.py
```diff
@@ -54,12 +54,13 @@ def search_for_nested_image_fields(
             # Get the corresponding field type with full generic info
             current_field_type = potential_field_types[idx]
 
-            # Check if it's a list or tuple generic type (e.g., list[ImageContent], tuple[ImageContent, ...])
-            if hasattr(field_specific_type, "__origin__") and field_specific_type.__origin__ in {list, tuple}:  # type: ignore[union-attr]
-                # Check if this container or its nested contents have images
+            # Check if it's a generic container type (list/tuple/dict).
+            # Example: list[ImageContent], tuple[ImageContent, ...], dict[str, ImageContent].
+            if hasattr(field_specific_type, "__origin__") and field_specific_type.__origin__ in {list, tuple, dict}:  # type: ignore[union-attr]
```
thomashebrard: Now that `dict` is included here, `check_generic_container_for_images` iterates `dict.__args__` indiscriminately, both the key type and value type. So a field typed `dict[ImageContent, str]` would be reported as
   an image field via the key type, and downstream template analysis would treat it as an image reference.

  In practice this is rare (Pipelex dicts are essentially always `dict[str, ...]`), but since we're adding the dict branch, let's make it robust. Could you tighten `check_generic_container_for_images` so that when
  `container_type.__origin__ is dict`, it only recurses into the value type, not the key?

and add a test

pipelex/core/stuffs/image_field_search.py
```diff
@@ -143,8 +145,8 @@ def check_generic_container_for_images(container_type: Any) -> bool:
     # Get the args (item types) from the generic
     container_args = getattr(container_type, "__args__", ())
     for arg_type in container_args:
-        # Check if arg_type is itself a generic (nested list/tuple) - recurse!
-        if hasattr(arg_type, "__origin__") and arg_type.__origin__ in {list, tuple}:  # type: ignore[union-attr]
+        # Check if arg_type is itself a generic container - recurse!
+        if hasattr(arg_type, "__origin__") and arg_type.__origin__ in {list, tuple, dict}:  # type: ignore[union-attr]
             if check_generic_container_for_images(arg_type):
```
thomashebrard: Top-level optionals are unwrapped before this helper runs, but optionals *inside* containers aren't. For `list[ImageContent | None]` or `dict[str, ImageContent | None]`, the inner `arg_type` is a `Union` /
  `UnionType`. Its `__origin__` is `typing.Union`, which isn't in `{list, tuple, dict}`, and `isinstance(arg_type, type)` is False — so the loop falls through and the collection is missed. `| with_images`
  validation would then reject a perfectly valid structure.

  This is technically pre-existing (the same gap exists for `list[ImageContent | None]` before your change), so if you'd rather scope this PR tightly, we can land it separately. But it'd be a clean addition here:
  when an `arg_type` is a `Union`/`UnionType`, iterate its non-`None` members and run the same checks on each.

thomashebrard:
Hey @matthewYang08 , thanks for your contribution and the fix, its great! but there is something missing:

Issue #521 specifically names `ListContent[ImageContent]` as the broken scenario, and your new `_is_image_collection_value` helper has a dedicated
  `ListContent` branch — but no test exercises it. Could you add a test in `tests/unit/pipelex/core/stuffs/test_structured_content_render_with_images.py` covering a `StructuredContent` field of type

thomashebrard:
You can also merge dev inside your branch, it is behind

pipelex/core/stuffs/structured_content.py
```diff
@@ -143,7 +145,10 @@ def render_with_images(
                 continue
             rendered = self._render_value_with_images(field_value, registry, text_format)
             if rendered:
-                parts.append(f"{field_name}: {rendered}")
+                if self._is_image_collection_value(field_value):
+                    parts.append(rendered)
+                else:
+                    parts.append(f"{field_name}: {rendered}")
```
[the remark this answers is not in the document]
matthewYang08: Fixed in the follow-up commits on this branch. Dict image collections now preserve semantic keys when rendered (for example `before: [Image N]` / `after: [Image N]`) instead of dropping them, and the regression tests cover that path.

pipelex/core/stuffs/image_field_search.py
matthewYang08: Addressed in the latest push as well. The helper now recursively unwraps inner `Union` / `UnionType` members, so optional image-bearing container items like `list[ImageContent | None]` and `dict[str, ImageContent | None]` are discovered correctly, with regression coverage added.

pipelex/core/stuffs/image_field_search.py
matthewYang08: Addressed in the latest push. `check_generic_container_for_images` now special-cases `dict` and only inspects the value annotation, and I added a regression test to ensure dict key annotations are not treated as image fields.

matthewYang08:
Sorry for the delayed follow-up on this PR.

I've now pushed a fresh update that addresses the review feedback and synced the branch with the latest `dev`.

What changed in the latest push:
- preserved semantic keys for dict-based image collections, so prompts keep context like `before: [Image N]` / `after: [Image N]`
- tightened `check_generic_container_for_images` so `dict[...]` only inspects the value annotation, not the key annotation
- added support for optional image-bearing container items inside collections (for example `list[ImageContent | None]` and `dict[str, ImageContent | None]`)
- added regression coverage for the dict key/value case, optional container-item discovery, and `ListContent[ImageContent]`
- merged the latest `dev` into this branch

Local verification completed on my side:
- `uv run ruff check`
- `uv run pyright`
- targeted unit tests for image field discovery and structured image rendering
- targeted integration coverage for prompt image extraction

When you have a moment, could you please take another look and kick off the remaining review/check workflow from your side if needed?

PR #1048, opened 2026-07-14
Release v0.39.0

lchoquel:
## Release v0.39.0

Bumps version from `0.38.0` to `0.39.0`.

### Highlights

- **From resolved library to typed clients: `pipelex resolve` and `pipelex codegen`** — `pipelex resolve` distills a bundle's whole closure into the normalized library crate — a flat, fully-qualified, fingerprinted snapshot — and `pipelex codegen types` projects it into TypeScript (`ts-zod`) or Python (`python-pydantic`, `python-structures`) artifacts. A trust chain of stamped headers and a `codegen.lock` makes drift detectable offline: `pipelex codegen check` verifies artifacts by pure hashing — no engine, no network, no API key.
- **Validation now fixes what it finds** — When an error has a deterministic safe fix, every surface says so: a 💡 suggested-fix line in the CLI output, a structured `suggested_fix` object in agent and API payloads. The new `pipelex fix bundle` / `pipelex-agent fix bundle` commands apply those fixes in place through a shared fix loop, with `--diff` to preview the changes without writing anything.
- **Smart Inputs: the signature decides the shape** — Inputs are now interpreted top-down against the pipe's declared input signature instead of bottom-up from their shape alone: a bare string becomes the declared concept (never silently degrading to generic `Text`), lists shape element-wise, bare URLs/paths resolve for `Image`/`Document` inputs, and generated input templates default to the light bare-value form.
- **New native concepts: `Date`, `Time`, and `YesNo`** — Built-in concepts for calendar dates (preserving source precision — no fabricated times of day), times of day, and boolean judgments, plus a `time` structure-field type completing the temporal triple. Every native now materializes from its pinned normative definition, so crate fingerprints byte-agree across implementations.

See `CHANGELOG.md` for the full entry, including the `### Added`, `### Changed` (several **Breaking** items), `### Fixed`, and `### Documentation` sections.

https://claude.ai/code/session_01KEVYPaF3KoVqmRFNdgN3mX


lchoquel:
## Codex review triage — 1 finding, confirmed and deferred

**P2 "Preserve extra working-memory entries from stdin envelopes" — ✅ confirmed (real regression), 📝 deferred out of this release by decision.**

Reproduced first-hand on this branch. An upstream sequence 's envelope carries its own inputs alongside its results (`root: ["text", "summary"]`), and `resolve_stdin_inputs` explodes all of them into inputs (`stdin_resolver.py:107-126`), so `InputShaper._check_input_names` (`input_shaper.py:136-142`) kills the run on the first undeclared name even though the declared one is present:

```
$ pipelex-agent run bundle chain.mthds --pipe prepare --inputs '{"text":"..."}' --dry-run --with-memory --format json > up.json
$ cat up.json | pipelex-agent run bundle chain.mthds --pipe classify --dry-run --format json
{"error": true, "error_type": "UnknownInputNameError",
 "message": "Input 'text' is not declared by this pipe. Declared inputs: 'summary'."}
```

It is a **regression vs v0.38.0** (`make_from_pipeline_inputs` took no `input_specs` and did no name check there; `input_shaper.py` does not exist in that tag), introduced with Smart Inputs (#1028). Worth adding to the bot's framing: the chained `run … --with-memory | run … --with-memory | run …` pattern is **taught in the shipped `mthds-plugins` skills** (`mthds-run/SKILL.md`, `shared/mthds-agent-guide.md`), so this is the CLI refusing its own documented output — not a hypothetical.

**Why it still ships:** it fails loudly with a precise, actionable message rather than corrupting data; workarounds exist (narrow the envelope, or pass `--inputs`); and the fix touches the core execution rails — and the MTHDS protocol, for `--runner api` parity — which is not release-day work.

**What landed instead** (8f6ace5, docs only):

- `wip/inputs/envelope-chaining-vs-d8.md` — the repro, the regression evidence, three fix options (recommended: an ignore-extras policy for *envelope-derived* inputs only, threaded along the rails `inputs_base_dir` already uses, keeping D8 strict for flat/file/inline inputs), and the open question of whether envelope chaining is local-runner-only or needs the same signal in the protocol.
- A **known-limitation sentence in the v0.39.0 changelog**, under the existing "Strict input validation (Breaking)" entry, so a user following the chaining docs is not blindsided.
- Follow-up recorded for the `mthds-plugins` skill docs, which teach the now-broken chain.

Thanks @chatgpt-codex-connector — good catch, and correctly scoped to `resolve_stdin_inputs`.

PR #1060, opened 2026-07-24
Generate concept structures in the sandbox for custom PipeFuncs

thomashebrard:
## Problem

A custom PipeFunc that returns a **non-native concept structure** (e.g. `price_quote` → `atlas_devis__PricedQuote`) must import that structure class. In the Daytona sandbox the class was neither present (the method's `structures/*.py` never travel) nor generated, so `from structures import ...` raised `ModuleNotFoundError` — which `_register_funcs_in_file` **silently swallowed** (`except ImportError: pass`), so the function never registered and the box failed with the opaque `Function 'price_quote' not found in registry`.

Simpler native-returning PipeFuncs were unaffected (they import no structure), which is why this only surfaced now.

## Fix

The `.mthds` is the single source of truth for concept structures, and the crate the box already receives carries every `ConceptBlueprint`. So the box now **regenerates `structures.py` from the crate** before registering the customer code (in `_run_transported_from_workdir`), forcing **domain-qualified** class names to match the runtime spelling the PipeFunc return-type validation checks (`concept.structure_class_name`). Structures are never stored or transported; a method-shipped file at the same path wins.

Also: `_register_funcs_in_file` now `log.warning`s a swallowed `ImportError` instead of `pass` — so a failed customer import surfaces its cause.

## Validation

- New regression test drives the full transported path with generated structures.
- `make agent-check` clean; targeted pipe_operators/libraries/system/codegen suites green.
- Verified against the real atlas method: the box generates `atlas_devis__PricedQuote` et al. and `price_quote` registers.

Note: this runs **inside the box** (`run_pipe_func_transported`), not on local direct runs (`run_pipe_func`), which are unchanged. Reaching the box also requires rebuilding the Daytona snapshot with this pipelex.

https://claude.ai/code/session_017LxfxhDrah899npndRcx7n

---
## Summary by cubic
Regenerates concept structure classes inside the Daytona sandbox from the crate so custom PipeFuncs that return non-native structures can import them and register. Also logs `ImportError` during function registration to surface real causes.

- **Bug Fixes**
  - Generate `structures.py` from the crate before importing customer code (transported path only); keep any shipped file instead of overwriting.
  - Force domain-qualified class names (e.g., `atlas_devis__PricedQuote`) to match runtime validation.
  - Log a warning on `ImportError` in `_register_funcs_in_file` instead of swallowing it.

- **Migration**
  - Affects only in-box runs; local `run_pipe_func` is unchanged.
  - Rebuild the Daytona snapshot to pick up this pipelex.

<sup>Written for commit aa1215619454ead6a8d504f3d2c2a180e9d0a61b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Pipelex/pipelex/pull/1060?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/pipelex/pipe_operators/func/direct_pipe_func_executor.py
+++ b/pipelex/pipe_operators/func/direct_pipe_func_executor.py
@@ -9,6 +9,8 @@ from typing import cast
 from typing_extensions import override
 
 from pipelex import log
+from pipelex.codegen.emitters.python_structures import emit_python_structures
+from pipelex.codegen.resolved_concepts import resolve_concepts_from_crate
 from pipelex.core.memory.working_memory import WorkingMemory
 from pipelex.core.stuffs.list_content import ListContent
 from pipelex.core.stuffs.structured_content import StructuredContent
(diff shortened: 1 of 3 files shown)


thomashebrard:
Superseded — re-opened from a fix/ branch to satisfy the gate-release branch-prefix guard (same commit aa12156).

PR #1071, opened 2026-07-29
fix(hub): complete the layer boundary — placement, so the predicate can say what it means

lchoquel:
## What this finishes

The hub split drew the runtime/interpreter boundary and made it enforceable. It shipped with a caveat repeated in **six** places: two interpreter packages could not be named in the closure test's predicate, so *"importing the Pipelex runtime loads zero interpreter modules"* was really *"zero modules from the packages we can name"*. This PR removes the caveat by fixing what caused it.

The measurement also turned up something the deferral did not know: one of the suspect packages is not an interpreter package at all, and it was hiding a module that loads the entire interpreter.

**Full guide, with every measurement and its reproduction recipe: [`wip/layer-placement-completion.md`](https://github.com/Pipelex/pipelex/blob/refactor/Layer-boundary/wip/layer-placement-completion.md).**

## Two findings, two commits, one theme — placement, not indirection

**1. `graph` was never an interpreter package, and being undeclared was hiding a live breach** (`463fe3a7b`)

`graph/graph_rendering.py` reached `interpreter_hub` through `pipeline.dry_run_pipeline` — 77 interpreter modules — with every gate green. The reason *is* the finding: both the layer rule and the transitive rule filter their candidates through `is_runtime_layer`, and `pipelex.graph` was not declared, so the module sat outside the rules' domain entirely.

> **An undeclared package is not neutral, it is unpoliced.**

Fixed by placement, along the boundary's own question — *do I need a loaded method?* The pure renderer (`GraphFormat`, `render_graph_from_spec`) stays in `pipelex.graph`; the bundle-driven half that must dry-run a method first becomes `pipelex.pipeline.bundle_graph_rendering`. `graph`, `tracing`, `observer` and `errors` — all measured to load zero interpreter modules — join `RUNTIME_LAYER_PACKAGES`.

**2. Four leaf models were filed in the wrong package** (`7e94bb3ac`)

They are what put `pipeline` and `pipe_run` in every runtime closure. Same remedy `mthds_parsing` used: move the leaf, then widen the predicate. Excluding it would only have recorded the problem.

| symbol | was | is now |
| --- | --- | --- |
| `SpecialPipelineId` | `pipeline.pipeline_models` *(module deleted)* | `system.job_metadata` |
| `PipeRunMode` | `pipe_run.pipe_run_mode` | `system.pipe_run_mode` |
| `PipeRunParamKey` | `pipe_run.pipe_run_params` | `system.pipe_run_param_key` |
| `PipeRunError` | `pipe_run.exceptions` | `core.pipes.exceptions` |
| `generate_graph_for_bundle`, `generate_view_for_bundle` | `graph.graph_rendering` | `pipeline.bundle_graph_rendering` |

Each destination is decided by what the symbol *is*. `PipeRunError` deliberately breaks the pattern: the runtime layer *subclasses* it (`PipeRunInputsError`, `OptionalValueAbsentError` in `core.pipes.inputs`), so it joins the pipe-error family whose base it is.

**3. Five vendor adapters were pulling the interpreter in through the public exceptions aggregate** (`08120e6c1`, found by the final review round)

Pre-existing, and a live counter-example to the very claim this PR strengthens: `pipelex.providers` — a package the declaration *already* vouched for — measured 15 interpreter modules. `providers/{anthropic,bedrock,google,mistral}/*_list.py` imported `MissingDependencyError` from `pipelex.exceptions`, the public all-errors aggregate that re-exports every interpreter package's errors, instead of from its definition site `pipelex.system.exceptions`.

It survived because it needed **two independent blind spots and had both**: the guard's rules only chase `interpreter_hub`, which an aggregate of exception classes never reaches; and the closure test's `providers.builtins` entry point misses these modules because the vendor plugins import them *inside a function* — a deferred import is invisible to the static graph and to the aggregator's runtime closure at once.

> **A module that re-exports across layers is a layer boundary with the sign filed off.**

Fixed at the import site per the repo's existing "direct full-path imports everywhere" rule, with `providers.anthropic.anthropic_list` added as an entry point (negative control run). `pipelex.test_extras` — clean, shipped, imported at boot, undeclared — is declared in the same commit, leaving every top-level package accounted for.

## What to check hardest

- **The declaration additions are pinned.** `test_the_measured_clean_packages_stay_declared` fails if any of the four names is dropped — negative control run. Without it, deleting an entry left the whole suite green, which is the exact regression this PR exists to close.
- **The property is unqualified now.** `INTERPRETER_PACKAGES` names every interpreter package, no exclusions. Measurement across all nine runtime entry points: **zero** `pipeline` / `pipe_run` modules. The closure test's negative control still fails with an offender list, so it is not passing vacuously.
- **The guard's negative control.** Re-adding the evicted import makes `make check-hub-layering` fail with `interpreter-hub-transitive`. Before the declaration it produced *no finding at all* — that contrast is the whole argument for declaring measured-clean packages.
- **`PipeRunError`'s destination** is a judgement call, argued in the plan. One earlier justification for it was **measured false and removed** during review (see below).

## Scale

~170 files, almost all mechanical. The substantive diff is `graph_rendering.py`, `bundle_graph_rendering.py` (new), `hub_layering_guard.py`, `system/pipe_run_mode.py` + `pipe_run_param_key.py`, the `PipeRunError` hunk in `core/pipes/exceptions.py`, `test_runtime_layer_import_closure.py`, and `docs/contribute/hub-layering.md`. Everything else is a one-line import swap — and a missed one is an `ImportError`, never a silent wrong resolution.

**No behavior changes.** Only where code lives and what the guard can see.

## Review record

Each phase was reviewed by an agent given only the diff — no plan, no rationale. Both rounds found real things:

- **Round 1** — the four new declarations were pinned by nothing (fixed, with negative control); the relocated test modules fell outside the documented targeted-test route (`tests/CLAUDE.md` row fixed).
- **Round 2** — two published doc snippets carried a now-dead import and would `ImportError` for anyone copy-pasting them; generated error pages were stale after `PipeRunError` changed module (`make gep`); and **a justification stated in the CHANGELOG, in `hub-layering.md` and in a drift ack was measured false**. The claim was that `system/exceptions.py` would make the config bootstrap load a run mode for nothing — but `import pipelex.config` already loads it via `runtime_hub → content_generator_protocol → assignment_models → cogt_run_params`. The clause was removed from all three and the drift ack re-recorded; the placement is unchanged and never needed that argument.

One question deliberately **not** answered, deferred to `wip/inputs/error-reference-grouping-axis.md`: the error reference groups by module path, so `PipeRunError` refiles from "Execution & runtime" to "Authoring & language". That is *consistent* — its two subclasses were already there, so the base now sits beside them — but whether that section is right for any of the three is a pre-existing product question.

## Gates

`make agent-check`, `make check-hub-layering`, full `make agent-test`, `make drift-check` — all green at each of the three commits. Both drift contracts (`cli-docs`, `hub-layering-convention`) reviewed for real and acked, with dogfood-log entries.

## Release-gated follow-up

Cross-repo consumers are on the old paths. Folded into the hub track's **already-open** sweep rather than opening a second one — see [`wip/hub/hub-split-tracker.md` → Third wave](https://github.com/Pipelex/pipelex/blob/refactor/Layer-boundary/wip/hub/hub-split-tracker.md#third-wave--the-layer-placement-track-refactorlayer-boundary). Measured: `pipelex-temporal` 37 sites, `cocode` 6, `pipelex-cookbook` 2, `pipelex-mistralai-workflows` 1. Both waves ship on the same release.

https://claude.ai/code/session_01EbQznTp6sHdxBuggUDZGS1


lchoquel:
@cubic-dev-ai please review — the check reported `skipping`, so I'd like a pass on this one.

Orientation, since ~170 files is mostly noise: the substantive diff is `pipelex/graph/graph_rendering.py`, the new `pipelex/pipeline/bundle_graph_rendering.py`, `pipelex/cli/dev_cli/commands/hub_layering_guard.py`, the new `pipelex/system/pipe_run_mode.py` + `pipe_run_param_key.py`, the `PipeRunError` hunk in `pipelex/core/pipes/exceptions.py`, and `tests/unit/pipelex/test_runtime_layer_import_closure.py`. Everything else is a one-line import swap.

Worth a hard look: whether the four relocated leaf models kept their exact semantics across the move, and whether `PipeRunError` landing in `core.pipes.exceptions` (rather than following the other three into `system/`) creates any import cycle — its two subclasses live one package below it in `core.pipes.inputs`.

lchoquel:
@greptileai please re-review — a fourth commit (`08120e6c1`) landed after the final review round found a live counter-example to this PR's own claim.

**What changed and why it matters:** `pipelex.providers` is a package the layer declaration *already* vouched for, and it measured 15 interpreter modules. Five vendor adapters (`providers/{anthropic,bedrock,google,mistral}/*_list.py`) imported `MissingDependencyError` from `pipelex.exceptions` — the public all-errors aggregate, which re-exports every interpreter package's errors — instead of from its definition site `pipelex.system.exceptions`.

It survived because it needed two independent blind spots and had both: the hub-layering guard only chases `interpreter_hub`, which an aggregate of exception classes never reaches; and the closure test's `providers.builtins` entry point misses these modules because the vendor plugins import them *inside a function*, hiding them from the static import graph and from the aggregator's runtime closure at once.

Worth checking specifically:

- the five one-line import swaps are equivalent (`MissingDependencyError` is defined in `system/exceptions.py`, not merely re-exported there);
- `providers.anthropic.anthropic_list` as a new `RUNTIME_LAYER_ENTRY_POINTS` member is a meaningful guard rather than decoration — I ran a negative control, reverting one import fails that entry point;
- `pipelex.test_extras` joining `RUNTIME_LAYER_PACKAGES` is correct (it ships, `pipelex.py` imports it at boot, and it measures zero).

The rest of the commit is documentation and cross-repo bookkeeping.

PR #1090, opened 2026-08-10
Plugins declare their layer through their entry-point group

lchoquel:
## Plugins declare their layer through their entry-point group

Part A of the kernel-layer plan (`wip/kernel/kernel-plugin-groups-and-distribution-plan.md`). The single `pipelex.plugins` entry-point group splits into **`pipelex.plugins.kernel`** and **`pipelex.plugins.interpreter`**, and `PLUGIN_API_VERSION` bumps to 4.

### The problem

A kernel-only boot loads zero interpreter modules — until you *install* an interpreter-side plugin. Discovery read one group unconditionally, so merely having our Temporal plugin on the path imported the whole interpreter into every boot, silently undoing the layering that v0.41.0 built and CI enforces statically. Static gates cannot see this: the contamination arrives through an installed distribution, not through an import in this tree.

### What changed

**A0 — the runtime layer is the kernel layer.** One name for one boundary. `KERNEL_BUILTIN_PLUGINS`, `KERNEL_LAYER_PACKAGES`, `is_kernel_layer`, the layer's prose, docs and closure tests. `runtime_bridge` keeps the word — it means the *orchestration venue*, a different sense. `runtime_hub` / `runtime_boot` still carry the layer's former name: they are pinned across a repo boundary (pipelex-transport's `ALLOWED_SURFACE`) and they move anyway in Part B, so renaming now would churn that boundary twice.

**A1 — the mechanism.** `build_registrar` takes a required `entry_point_groups`, symmetric with the existing `builtin_plugins` injection; the composed defaults live beside the built-in manifests they belong with. Three gates, each mutation-tested:

- **A kernel-only boot queries only the kernel group**, so an interpreter-side plugin's module is never imported.
- **Menu-tier cross-check** — a kernel-group plugin that registers an orchestrator, a bundle validator, a PipeFunc executor, or an interpreter-layer hub slot fails loud at register time. One-directional on purpose: an interpreter-group plugin may contribute kernel-tier capabilities, because a symmetric rule would reject our own Temporal plugin (interpreter-side, yet contributes an HTTP-error mapper) while rejecting nothing dangerous.
- **The retired group is a startup error**, never a silent nondiscovery — the expensive symptom is a capability simply absent.

`HubSlot.is_interpreter_layer` uses an exhaustive match, so a new slot cannot be merged without classifying it.

**A3 — the gate that needed a real distribution.** `tests/unit/pipelex/plugins/test_installed_plugin_group_isolation.py` builds an interpreter-group plugin as an actual installed dist in a temp site-dir, whose module writes a sentinel on import, and boots a subprocess through `RuntimeBoot.make()` passing **no** `entry_point_groups` — so the production default is what is under test, not one the test supplies. A both-groups discovery in the same subprocess proves the fixture is not inert.

### Scope worth stating

This governs which plugins are **discovered** and what they may **register** — not what a plugin's module *imports*. A kernel-group plugin with a module-scope interpreter import still pulls it in; only an import-closure rule would prevent that. Recorded in the changelog and in `wip/kernel/plugin-group-split-deferred-items.md` (D-2).

### Breaking

- `build_registrar` gained a required `entry_point_groups`. Any host resolving plugin contributions itself — notably one reading `get_http_error_mappers()` — must pass it, and fails at *import* when the call sits at module scope.
- The `pipelex.plugins` entry-point group is retired.
- `PLUGIN_API_VERSION` → 4.
- `begin_plugin(group=...)` is required, not defaulted.

### Ships with, but cannot go first

Released pipelex reads only the retired group, so the five downstream repos' CI would install it and go red for a reason no code change fixes. Order: **land this**, cut the release, then push the five branches (pipelex-temporal, pipelex-mistralai-workflows, pipelex-daytona-sandbox, pipelex-cookbook, pipelex-api), open their PRs, and bump each pin in the same window. Each already holds its commit locally.

`pipelex-api` is the one found late: it declares no entry point, so a sweep keyed on entry-point declarations returned zero hits on it — but it *calls* `build_registrar` at module scope. Import-time `TypeError`, dead container. Confirmed red-green against this branch installed editable.

### Verification

`make agent-check`, full `make agent-test`, and `make drift-check` all green on the tip. Two drift contracts re-acked. A `/code-review` fan-out over the whole of Part A ran at the checkpoint; its confirmed findings are fixed in the last commit, one finding was rejected on empirical evidence (pyright's `reportImportCycles` flags TYPE_CHECKING-only cycles, so `PluginGroup` stays in its own module), and the judgment calls are deferred in `wip/kernel/plugin-group-split-deferred-items.md` — which also carries three open questions for @lchoquel.

---
## Summary by cubic
Split plugin discovery by layer and rename the “runtime layer” to “kernel layer.” Plugins now publish under `pipelex.plugins.kernel` or `pipelex.plugins.interpreter`; kernel-only boots read the kernel group only to keep interpreter code out.

- **New Features**
  - Two entry-point groups: `pipelex.plugins.kernel` and `pipelex.plugins.interpreter`; publishing under retired `pipelex.plugins` raises `RetiredPluginEntryPointGroupError`.
  - Reject a plugin declared in both groups with `PluginDeclaredInMultipleGroupsError` (scoped to the groups a boot reads).
  - Enforce layer at registration: a kernel-group plugin cannot register interpreter-tier capabilities; violations raise `PluginLayerViolationError`. `HubSlot.is_interpreter_layer` is exhaustive and clarifies `TASK_MANAGER` as kernel-tier.
  - `build_registrar` requires `entry_point_groups`; defaults live with built-in manifests. Adds `PluginGroup`. CLI `pipelex plugins list` shows a “Group” column. Docs updated and include the kernel distribution footprint (B0).
  - `PLUGIN_API_VERSION` bumped to 4.

- **Migration**
  - External plugins must publish under `pipelex.plugins.kernel` or `pipelex.plugins.interpreter` (not `pipelex.plugins`) and set `targets_api = 4`.
  - Hosts must pass `entry_point_groups` to `build_registrar(...)` and supply `group=` to `begin_plugin(...)`.

<sup>Written for commit 43f9101e4f5166c9ca09371becf1bdd55f0548f6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Pipelex/pipelex/pull/1090?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lchoquel:
@greptileai review

@cubic-dev-ai review

lchoquel:
Both findings fixed in `bf2317372` — thanks, and the second one had a sharper edge than the report claimed.

**The isolation gate's exact-list assertion** (@chatgpt-codex-connector + @cubic-dev-ai, same finding): correct, and the affected population is exactly the developers most likely to run this suite. Now a membership check. The property is "our fixture was discovered"; "nothing else was" is a fact about the machine, and the subprocess deliberately keeps the ambient site-packages.

**The dual-group declaration** (@cubic-dev-ai): valid, and it is not only an opaque-message problem. Every keyed menu method does collide on the second `register()`, but `add_teardown` is a bare `append` with neither a key nor a tier guard — so a plugin declared under both groups that contributes only a teardown callback registers twice **silently** and runs its teardown twice at shutdown. Discovery now raises `PluginDeclaredInMultipleGroupsError`, scoped to the groups actually read: a kernel-only boot sees such a plugin once and is unaffected, so probing the group it deliberately does not query would buy a diagnostic for a boot with nothing to diagnose.

Both guards mutation-tested — removing the check kills the dual-group test; keying the dedup on the name alone kills the single-group test written to catch exactly that over-reach. `make agent-check`, full `make agent-test` and `make drift-check` green; the new error class regenerated its identity snapshot and reference page.

lchoquel:
Second pass — I had truncated the comment list and triaged only three of six. The remaining three (@cubic-dev-ai) are all valid and fixed in `3e4e68126`.

**`contract.py` invariant** (confidence 10 — deservedly): the docstring stated "a plugin belongs to exactly one layer… a capability that needs both is two plugins", then three sentences later in the same paragraph said ours contributes both. An author reading the bolded invariant first would split a valid mixed plugin or publish it under the wrong group. The invariant is really *a plugin belongs to the highest tier it contributes to* — which is what the registrar enforces — so it now says that, with the one-directional rule and its reason stated separately from the declaration rule.

**The orchestrator-as-kernel-tier example**: right, and it contradicted `add_orchestrator`'s own guard three lines below the sentence. Each capability is now labelled with its tier ("an orchestrator (interpreter-tier) *and* an HTTP-error mapper (kernel-tier)"). I swept for other copies — the three elsewhere in the tree already attached "kernel-tier" to the mapper correctly, so only these two were muddled.

**The unindexed A2 note**: added to the working-notes index.

Worth noting these three were the same class of defect as the four I fixed at the checkpoint review — docs describing a rule stricter or different from the one the code enforces — and `contract.py` is the one an external plugin author actually reads, so it was the most consequential of the set.

`make agent-check`, the plugin suite and `make drift-check` green.

PR #1100, opened 2026-08-13
Per-node token usage, cost and model attribution on the run GraphSpec

thomashebrard:
## What

A run graph and a cost report describe the same run, built from the **same** trace-event stream — and every usage event already names the node it belongs to. The assembler was throwing that link away (`elif isinstance(event, UsageReportEvent): pass`), so no consumer could answer *which pipe spent the money*.

This keeps the link. `NodeSpec.usage` and `GraphSpec.usage` now carry per-node inference usage, rolled up the containment tree.

## The shape

`NodeUsageSpec` — own half and a `subtree_*` half (this node plus every descendant):

- `inference_calls`, `rated_inference_calls`
- `cost`, split into `cost_input` / `cost_output`
- `nb_tokens_by_category`, `total_tokens`
- `by_model[]` — name, id, `model_type`, calls and cost of every model that **actually ran**

`GraphUsageSpec` on the graph: `total` (every call the run reported) and `unattributed` (calls naming no live node — surfaced rather than dropped, so the graph's total can never silently disagree with the cost report's).

## Five invariants, stated on the model

They are the contract, and each has a named test:

1. `usage is None` is **all-or-nothing across a graph** — no usage reported anywhere. Once one event is seen, *every* node carries a spec, zeroed where nothing ran. It never distinguishes "this node was not measured" from "that one was".
2. `cost is None` <=> `rated_inference_calls == 0`. "Made no call" and "made only unrated calls" both land here; `inference_calls` tells them apart. **Never render it as `$0.00`.**
3. `inference_calls > rated_inference_calls > 0` means `cost` is a **lower bound**. A UI showing it as complete is silently wrong about money.
4. `total_tokens` is input_joined + output — **not** the sum of `nb_tokens_by_category` (`input_cached` is a subset of `input`; summing double-counts).
5. `cost_input + cost_output == cost`, and all three go `None` together.

## Why `by_model` exists

A GraphSpec named a model in two places and neither was the outcome: the blueprint holds the authored choice (`$writing-factual`), and `execution_data.resolved_model` holds the handle `LLMSetting.model` carried — frequently still an unresolved alias, since aliases resolve at inference time. Across the fixture corpus **48 of 159** such values were bare `@aliases`.

`by_model` reads the usage records instead, so it survives alias resolution, deck defaults, and any fallback or retry. It is a **list** because one node routinely uses several models — a `PipeLLM`'s text pass and object pass resolve separately.

`model_type` rides along as the discriminator a consumer needs **before displaying token counts**: extract, search and img_gen are billed *per request*, and pipelex encodes that price by putting `1_000_000` in each token category (rates are per-million). A one-page extract reports 2,000,000 "tokens" — a scaled request counter, not a measurement. A controller's subtree can blend the two, which is why the flag is per model rather than per pipe type.

## Design notes

- **Attribution resolves in pass 2, not in the handler.** A `UsageReportEvent` can be read *before* the `PipeStartEvent` of the node it names — different workers, stream ordered by `(workflow_id, sequence)` — so "is this node real?" is only answerable once every event is in. Tested.
- **One cost engine.** `compute_tokens_usage_costs` now returns the components and `compute_tokens_usage_cost` is a thin wrapper, so the split and the headline cannot disagree. No cost math is re-derived anywhere.
- **Malformed parentage is tolerated, not fatal.** A cycle is broken by an in-progress guard and logged; a dangling parent becomes a root instead of a `KeyError`. Both reachable from a partial cross-worker read. The rollup is memoized with an explicit stack — O(nodes), independent of the recursion limit.
- **The in-process `GraphTracer` deliberately does not carry usage.** Its GraphSpec is discarded at all three call sites and `GraphTracerProtocol` has no method to receive it. `test_assembler_equivalence` asserts that divergence explicitly rather than normalizing it away, so plumbing it in later fails loudly.

## Breaking

`NodeSpec` and `GraphSpec` are `extra="forbid"`, so an older pipelex rejects the new JSON. Per workspace policy there is no transition window.

Related: GraphSpec still has no version field despite the docs calling it versioned — pre-existing, and this change makes it more visible. Captured as a TODO.

## Docs

New Per-node Usage Attribution page under `docs/under-the-hood/`, covering the two projections of one event stream, the usage states, and the model ladder. `cost-tracking.md`, `execution-graph-tracing.md` and `tokens-usage-wire-records.md` updated.

## Known defect this surfaced, NOT fixed here

The `1_000_000` sentinel is not confined to the graph — it rides the `/execute` API's `tokens_usages` records and the CLI cost table's token columns, and `AggregatedCosts.total_nb_tokens` sums it. Any client computing token usage from the API gets nonsense once a method touches extract or search. The fix is a `billing_unit` discriminator on the usage record; that is an API-contract change and needs its own plan.

## Verification

`make agent-check` clean. `make agent-test`: 9665 passed. The 6 failures are all `test_bedrock.py` network/credential integration tests, unrelated to this change. Verified end to end against 32 live pipeline runs.

---
## Summary by cubic
Attributes per-node token usage and cost to run `GraphSpec` nodes (with input/output split and per-model attribution), resolves recorded models to concrete handles so dry and live runs agree, and makes dry-run condition branching deterministic. Adds a docs page with mermaid diagrams for per-node usage attribution.

- `NodeSpec.usage` and `GraphSpec.usage` report call counts, `nb_tokens_by_category`, `total_tokens`, `cost`, `cost_input`, `cost_output`, and `by_model[]` (with `model_type`), plus subtree rollups; numbers map directly from `CostRegistry.aggregate_costs`. Invariants: cost is null iff unrated; `cost_input + cost_output == cost`; `total_tokens` equals input_joined + output.
- Attribution runs post-ingest; usage without a node is stamped `UNATTRIBUTED_NODE_ID` and surfaced in `GraphUsageSpec.unattributed`; run-level totals are in `GraphSpec.usage.total`.
- `execution_data.resolved_model` now uses `concrete_llm_model_handle` to report a fully resolved model handle; waterfalls pass through unchanged; unknown references are returned as-is.
- Dry-run determinism: `PipeCondition` sorts `pipe_dependencies()` to stabilize branch order and resulting graphs. Tests scope their patching of `pipe_dependencies()` to avoid affecting library loading.
- Docs: new Per-node Usage Attribution page and updates across related docs; diagrams switched to mermaid for clarity.

**Migration**
- `NodeSpec` and `GraphSpec` are `extra="forbid"`. Update consumers to read `usage`, treat null cost as unrated (do not render $0.00), show mixed rated/unrated as a lower-bound cost, and include `GraphUsageSpec.unattributed` in run totals.

<sup>Written for commit b530ee2357e83bac547b12e7ea564d65561b92b9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Pipelex/pipelex/pull/1100?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


docs/under-the-hood/per-node-usage-attribution.md
```diff
@@ -0,0 +1,140 @@
...
+              ┌──────────┴──────────┐
+              ▼                     ▼
+      GraphSpecAssembler       UsageAggregator
+      (node_id kept,           (node_id dropped,
+       calls folded)            one record per call)
+              │                     │
+              ▼                     ▼
+   NodeSpec.usage /            PipeOutput.tokens_usages
+   GraphSpec.usage             → cost table, CSV, API wire
+```
```
lchoquel, line 48: our md docs can display mermaid, it could be nicer

lchoquel (dismissed):

PR #1122, opened 2026-08-18
feat!: storage_scope — the host says where a run's bytes go

thomashebrard:
The runtime half of the hosted org-scoped storage change. Paired with Pipelex/pipelex-server#29.

## The problem

Storage keys were built from `user_id`:

| Writer | Key |
|---|---|
| delivery executor | `{user_id}/{key_prefix}{run_id}` |
| generated content | `{primary_id}/{secondary_id}/…` = `{user_id}/{run_id}/…` |
| input normalizer | a flat top-level `normalized/{uuid}` |

That ties **where the bytes live** to **who uploaded them**. Fine until two people share the work: the second one asks for a file the first attached and is refused, because the only thing standing between them is a string comparison on path segment 0. A host serving teams cannot express *"these two people may read the same file"* in that layout at all.

And the flat `normalized/` prefix was shared by every run of every tenant.

## `JobMetadata.storage_scope`

One opaque, host-supplied prefix. The runtime composes `assets/`, `results/` and `payloads/` onto it and never parses it.

**The runtime does not learn what a tenant, an organization or a method is, and must not.** Threading the host's own concepts through the transport was considered and rejected — `method_id` is a hosted catalog concept with no meaning in an MIT-licensed runtime, and `uri_format`'s placeholder set is closed by design. The hosted platform fills the scope with `<org_id>/<method_id>/<run_id>`; a single-user deployment can pass `<user_id>/<run_id>`; a local one passes `local`.

**Validated at construction, on the type.** The value becomes a storage key prefix, so a `..` or a leading slash in it is a traversal out of the tenant's namespace. A `JobMetadata` that exists is one whose scope is safe, which makes every key derived downstream safe *by construction* rather than by each call site remembering. This is not optional cleanliness: the Temporal payload codec used to sanitize `user_id` and `pipeline_run_id` per-segment, and collapsing them into one slash-bearing string makes that sanitizer unusable — without the validator, this change would have silently deleted an existing traversal control.

## Identity becomes required, and `"anonymous"` goes back to being a tracing label

`pipeline_run_setup` did `user_id = user_id or OTelConstants.DEFAULT_USER_ID`. That constant is the string `"anonymous"` — a **tracing** placeholder for a span with no known caller. It leaked into the identity path and became a storage key prefix, so every run without an authenticated caller wrote into one shared `anonymous/` namespace where each tenant could read the others' outputs.

The fallback is what made it silent: a missing identity looked exactly like a present one. `user_id` and `storage_scope` are now required parameters of `pipeline_run_setup` and `prepare_pipe_job`, so the same omission is a `TypeError` at the call site instead of a shared prefix in production. The constant survives for telemetry only.

Two callers pass explicit, greppable values rather than inheriting a default:

- a **dry run** passes `DRY_RUN_STORAGE_SCOPE` — it provably stores nothing, but needs a value for a required field, and if `dry-run-no-storage` ever shows up as an S3 prefix then a dry run stored something and that is the bug to chase;
- **`PipelexMTHDSProtocol`** defaults to `LOCAL_USER_ID` / `LOCAL_STORAGE_SCOPE`, because a run on somebody's own machine genuinely has one user and no tenancy.

A constructor default stating a true fact at the boundary where it is true is a different thing from an `or` fallback deep in the call stack — and a multi-tenant host cannot reach these by omission, because the seam it calls requires both explicitly.

## Breaking

| Surface | Change |
|---|---|
| `JobMetadata` | required `storage_scope` |
| `PipelexPipeRunInput` | required `storage_scope`; `user_id` no longer nullable |
| `pipeline_run_setup`, `prepare_pipe_job`, `PipelexKernel.make` | `user_id` + `storage_scope` required |
| `DeliveryExecutor.execute` / `_store_results` | `user_id` → `storage_scope` |
| `GeneratedContentFactory` | `(primary_id, secondary_id)` → `storage_scope` |
| `normalize_data_urls_to_storage` | takes `storage_scope` |
| `uri_format` | `{secondary_id}` removed; default is `{primary_id}/{hash}.{extension}` |

**A config still naming `{secondary_id}` is refused at boot**, by name, with the supported set listed. **No ledger entry**, and the gate agrees: it reports the change as additive — *"the fingerprint moved additively … Nothing breaks and no entry is needed"* — because the supported-placeholder set is a `ClassVar` the fingerprint cannot see. And per `docs/migration-ledger.md`, a free string's value domain can never be proven stale from the schema, so a `remap_value` there could only be `unsafe` (reported, never applied). The boot error already names the exact fix. Goldens regenerated with `make umig`; `cmig` and `cl` both pass.

## Tests

`make agent-check` clean. `tests/unit`: **9328 passed**, 2 failed.

Both failures are **pre-existing and environmental** — `test_runtime_boot_closure` and `test_kernel_boot_contract`, which fail identically on a stashed clean tree in this checkout (a stale global `~/.pipelex/` config: *"Extra forbidden fields: 'pipelex', 'cogt'"*). Neither touches storage. CI should confirm.

New: `tests/unit/pipelex/system/test_storage_scope.py` — the construction-time guard, including traversal in every position, empty and four-segment scopes, percent-encoded traversal, and the explicit record that the validator checks **shape, never ownership** (`tenant2` is a valid scope even when the caller is `tenant`; refusing it needs to know who is calling, which this layer deliberately does not).

## Release

This must be published before Pipelex/pipelex-server#29's phases 4–5 can move: the server pins `pipelex==X.Y.Z` at five sites and cannot consume an unreleased runtime. Accumulated under `## [Unreleased]` per the versioning rule — no version bump in this PR.

---
## Summary by cubic
Host-controlled storage via a required, validated storage_scope. Previously keys were derived from user_id/pipeline_run_id and inputs used a global normalized/ prefix; now the host supplies an opaque scope and the runtime writes under {storage_scope}/assets|results|payloads to prevent cross-tenant leakage. Also forbids binding the telemetry placeholder "anonymous" as an identity.

- JobMetadata now requires storage_scope and validates traversal (.., leading /, percent-encoded traversal). "anonymous" remains tracing-only; a guard test fails if it is ever bound to user_id or storage_scope.
- API: pipeline_run_setup, prepare_pipe_job, PipelexKernel.make, and PipelexPipeRunInput require non-null user_id and storage_scope. DeliveryExecutor.execute/_store_results and GeneratedContentFactory take storage_scope; delivery artifacts land under {storage_scope}/results/….
- Input normalization normalize_data_urls_to_storage(..., storage_scope=...) writes under {storage_scope}/assets/ (was top-level normalized/).
- Config: uri_format supports only {storage_scope}, {hash}, {extension}; any use of {primary_id}/{secondary_id} is refused at boot. Shipped defaults use {storage_scope}/{hash}.{extension}.
- Safe defaults: dry runs use DRY_RUN_USER_ID and DRY_RUN_STORAGE_SCOPE; local runs default to LOCAL_USER_ID and LOCAL_STORAGE_SCOPE.
- Review note: DeliveryExecutor composes {storage_scope}/{key_prefix}results/…; a new test pins that the results/ leaf appears exactly once even if callers include "results" in key_prefix.

Required migration actions:
- Provide both storage_scope and user_id at all call sites creating runs or invoking the kernel, including payloads (PipelexPipeRunInput).
- Replace any {primary_id}/{secondary_id} placeholders in storage uri_format with {storage_scope}.
- Pass storage_scope through input normalization, content generation, and delivery storage paths.
- Do not bind OTelConstants.DEFAULT_USER_ID ("anonymous") to identity or storage_scope; use explicit values.

<sup>Written for commit 857af35c1205f9ed2208417eb2ffc983d0decda7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Pipelex/pipelex/pull/1122?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


thomashebrard:
Superseded by #1129.

Not a content change — same commits plus the fixes for the review findings below. The branch prefix `feat/` is refused by `gate-release` (it allows `fix|feature|refactor|chore|docs|ci-cd|changelog|codex`), and renaming a branch closes its PR on this repo, so the work moved to `feature/storage-scope` rather than being renamed under this one.

The eight threads here were all triaged; five were real and are fixed in #1129, each pinned by a test verified to fail without the fix. Details in that PR's description.

PR #1159, opened 2026-08-28
Stop four wrap serializers from erasing their serialization JSON Schema

lchoquel:
## What

Four models published an empty serialization-mode JSON Schema — `{"type": "object", "additionalProperties": true}`, with no `properties`, no `required`, no closed shape:

- `ConceptBlueprint`
- `ConceptStructureBlueprint`
- `InputSlotBlueprint`
- `ConstructBlueprint`

Pydantic derives a model's **serialization** schema from its `@model_serializer`'s return annotation and prefers it over the schema it would generate, so the `-> dict[str, Any]` on each wrap serializer replaced the real shape with an opaque object. Dropping the annotation restores the shape and leaves `model_dump()` byte-identical — the same fix `mthds` 0.11.1 applied to the input-form field descriptors, which were hit by the identical bug after PR #1154 moved them upstream.

The models never loosened: validation-mode schemas stayed complete and `extra="forbid"` still holds. Only the *published description* was gone, and it was gone exactly where a server generates response-model schemas, since FastAPI always generates responses in serialization mode. The visible symptom is in `pipelex-api`'s `docs/openapi/pipelex-api.openapi.yaml`, where two of these blueprints regressed to opaque objects on a `pipelex` bump; that artifact is regenerated mechanically, so the fix has to be here. It regains its field-level detail on the next pin bump, with nothing to do in that repo.

`ConstructBlueprint` was not in the report — a whole-package sweep found it, same cause and same wire surface. `ModelReference.serialize_to_raw_string -> str` is deliberately left alone: a precise annotation is desirable, and `{"type": "string"}` is its correct published schema.

## Why nothing caught it

No call to `model_json_schema()` anywhere in `pipelex/` asks for serialization mode — the MTHDS schema generator asks for `mode="validation"` explicitly — so the suite was green while the wire surface was empty. That silence is why the bug class recurred in two repos, and it is why the deliverable here is the guard, not the four one-line edits.

`tests/unit/pipelex/test_serialization_json_schema.py` imports every `pipelex.*` module, walks `BaseModel.__subclasses__()`, and fails any model that is opaque when serialized but **not** when validated. Testing that asymmetry rather than the shape itself is what lets it run without an allowlist: a model that is honestly open — `extra="allow"` with no declared fields, or a `RootModel[dict[str, Any]]` — is opaque in both modes and passes, because an open dict really is its shape. An allowlist would be the wrong instrument anyway, since it would have to be extended by the very person adding the next erasing serializer. A second, sharper test compares the two modes' property lists on the four fixed models, catching a partial erasure the sweep would wave through, and an anti-vacuity test asserts the walk imported modules with zero failures and pins the one model pydantic cannot describe at all.

## The rule that caused it

`python_standards.md` § "Always Use Type Hints" says every function return must be typed — that rule is why the annotation was written here and upstream. It now carries a scoped carve-out for pydantic serializers, mirrored into the workspace repo's copy in a separate commit, and each fixed serializer keeps a docstring paragraph saying the return is deliberately unannotated. Without that, the next reader puts it back, correctly following the rule.

## Verification

- `make agent-check` and the full `make agent-test` are green; `make rules` regenerated `AGENTS.md` and `make check-rules` passes.
- The new guard was mutation-tested: re-adding the annotation on `ConceptBlueprint` reddens both tests, naming the model and the remedy.
- `derived/mthds_schema.json` does not move, as expected — it is generated in validation mode.

## Note on the branch

This branch sits on top of `release/v0.54.0` while its PR to `main` is still open, so it carries the release commit as a ride-along until that merges back.

Closes L-260824-52d6fe

https://claude.ai/code/session_01UGJGZZAJ7rzcKuts46okQo

---
## Summary by cubic
Restores the serialization-mode JSON Schema for `ConceptBlueprint`, `ConceptStructureBlueprint`, `InputSlotBlueprint`, and `ConstructBlueprint`. Pydantic was using each wrap serializer's `-> dict[str, Any]` return annotation as the model's serialization schema, publishing an opaque `{"type": "object", "additionalProperties": true}` instead of the real shape. Dropping the annotation restores the fields and leaves `model_dump()` byte-identical; the same fix landed in `mthds` 0.11.1 for the input-form descriptors. FastAPI always generates response schemas in serialization mode, so downstream OpenAPI artifacts regain field-level detail on the next `pipelex` bump. No migration needed: validation-mode schemas were unchanged and `extra="forbid"` still holds. This is the same bug class that recurred upstream, so the deliverable is the guard.

**New Features**
- Adds a sweep test that imports every `pipelex.*` model and fails any model that is opaque when serialized but not when validated, so future erasing serializers get caught without an allowlist.

Closes L-260824-52d6fe

<sup>Written for commit 00b63bb2e681da01f9285db6639a96d04bb20a99. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Pipelex/pipelex/pull/1159?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lchoquel:
@cubic

lchoquel:
@codex

PR #1186, opened 2026-09-02
Output-form descriptor and a payload schema on the output contract

thomashebrard:
## What

The validate report gains a third artifact, `output_form`, beside `pipe_io_contracts` and `input_form`, and the output contract gains a `json_schema` where it carried none.

Both close the one real asymmetry between a pipe's two halves: an input had a presentation view and a payload schema; an output had neither. A tool registry had no return type, and a result renderer had to work out a payload's shape by inspecting it — which is guessing, and it was wrong in production (a Markdown report rendered as plain text, a list of short strings rendered as a stack of cards, a `native.Composite` rendered as a JSON blob).

## How

**The builder is a projection, not a new derivation.** `InputFormDeriver.derive_concept` — already called for every nested concept-typed field of every input — is what describes an output, because an output is a concept reference exactly like an input is. It stamps no slot facts (`presence` and `gating` belong to an input slot; an output is not one), and it iterates the same `result.pipes` as the two builders beside it, so all three artifacts share one key set by construction rather than by intent.

**The plural wrap is the one place the builder does real work.** Plurality is not on the concept — `concept_ref` is the element with the multiplicity suffix stripped on both sides — so `derive_concept` cannot know it. `build_output_form` reads `pipe.output.multiplicity` and wraps a plural output as a `list` node whose `item` is the element.

**The output schema is the PAYLOAD's, which is where it departs from its input twin.** An input's schema states what a caller *sends*, so a plural slot's is a bare array. An output's states what *comes back*, which is a `StuffContent` — the single arm goes through the same `render_stuff_spec` call the inputs do, and the plural arm is `ListContent[structure_class]`, asked of pydantic rather than hand-built.

## Two fixes that came out of using it

- `native.Anything` on either side of a contract crashed `_render_output_schema`; it now yields the empty schema, which is what "anything" means.
- `HtmlContent`'s JSON rendering stated members the model does not have.

## Notes

Requires `mthds` 0.12.0. A new corpus bundle covers outputs specifically — flat and nested structures, both plural arms, an optional output, a wrapping native scalar and a file-bearing one. It exists because every bundle before it resolved to `Text`, so the corpus could only ever capture one output descriptor and one output schema.

Consumed downstream by `mthds-js`/`mthds-python` (protocol types), `pipelex-api` (the `views: ["output_form"]` opt-in), both SDKs, `@pipelex/mthds-form`'s `StuffViewer` and `@pipelex/mthds-ui`'s graph detail panel.

https://claude.ai/code/session_019VJ89z3Gza9CW25pQombFK


thomashebrard:
Pinned `mthds` to the mthds-python branch SHA so CI resolves the code this branch imports:

```toml
[tool.uv.sources]
mthds = { git = "https://github.com/mthds-ai/mthds-python", rev = "570cf7c657ac2bcdac68055efb18d8d03c52e5b9" }
```

That SHA is mthds-ai/mthds-python#59's HEAD.

**Why it was needed, and the release problem it exposes.** This branch imports `mthds.protocol.output_form`, and no published `mthds` has that module. `dependencies` says `mthds==0.12.0` and that version *is* on PyPI — but its wheel ships `input_form.py` and nothing beside it. The number was consumed by a build predating the descriptor, so a resolver that trusts the range installs a package missing the module, and the failure arrives as an ImportError at collection rather than as a version conflict anyone would read as one. The lockfile was in fact resolving 0.11.1, a further version behind.

**This pin makes CI honest; it does not make a release correct.** `[tool.uv.sources]` is dev-and-CI metadata — it is not part of the published dependency, so `pip install pipelex` still resolves `mthds==0.12.0` and still gets the wheel without the module.

Before merge, both of these in one change:
- move `dependencies` to whatever version mthds-python actually publishes the output-form change as (**0.12.0 cannot be reused** — PyPI forbids re-uploading a version), and
- **delete the `[tool.uv.sources]` block**, then re-lock.

Pinned by SHA rather than branch name so resolution cannot drift under a force-push while this PR is open.

lchoquel (dismissed):

thomashebrard:
**Git pin removed — this now depends on the published `mthds==0.13.0`.**

`[tool.uv.sources]` is deleted and the dependency is an ordinary registry version again, so `pip install pipelex` and CI resolve the same thing. The earlier pre-merge checklist on this PR is discharged.

0.13.0 rather than 0.12.0 because that number was already taken by a build predating the descriptor — its wheel carries `input_form.py` and nothing beside it, and PyPI does not allow re-uploading a version. I pulled the 0.13.0 wheel and confirmed `mthds/protocol/output_form.py` is in it before switching.

One note for anyone re-locking: a plain `uv lock` reported `no version of mthds==0.13.0` from a cached index, which reads like the release does not exist. `uv lock --refresh-package mthds` resolves it.

`make agent-check` passes against the published package, and the branch is current with `dev`.

lchoquel (approved):

PR #1206, opened 2026-09-16
feature/Log-sink-seam · L-260916-aa6e58 — sink registry, json otlp console

lchoquel:
Where log records go is now a sink the plugin registrar registers and one config key selects, with json, console and otlp sinks built in, a holding handler that keeps what a boot says until the sink is installed, and the rich-or-poor log mode retired with a migration entry. Stacked on feature/Log-fields-context, which it needs for the fields and context the sinks carry.

Closes L-260916-aa6e58

https://claude.ai/code/session_016gUk6TEF3BG4qc4qeaSfvy

---
## Summary by cubic
Log records are now routed through a log sink selected by one config key, `runtime.log.sink`, with `console`, `json`, and `otlp` built in and external sinks registerable through the plugin registrar. Records emitted before the sink is installed are held and replayed through it; the rich-or-poor log mode is retired, and an unregistered `sink` token stops boot naming the registered ones. Closes L-260916-aa6e58.

**Migration**

- `pipelex-config@5` is `safe`: it deletes the four retired keys (`log_mode`, `poor_loggers`, `generic_poor_logger`, `is_console_logging_enabled`), so older files at their old defaults migrate in one run.
- No migration operation writes a replacement value, so a file that set `log_mode = "poor"` or `is_console_logging_enabled = false` must be edited by hand to `sink = "json"` and `pipelex = "OFF"` under `[runtime.log.package_log_levels]`.

**Bug Fixes**

- `doctor` reports an unregistered `sink`, a plugin that stops discovery, or a sink that fails to install as rows, pins its report console to stderr, and never claims a fallback sink that was not installed.
- `include_exception` carries the exception on `exc_info` instead of in the message, so each sink renders it in its own shape.
- The holding handoff respects the sink handler's level, routes unrenderable records to `handleError`, marks forwarded records, and keeps every held record above the configured level, so a boot that dies before the sink arrives loses no trail.
- Colliding names — `data`, a sink's own keys, `_pipelex_forwarded`, and JSON/OTLP reserved attributes — are carried under a `field_` prefix, and non-finite floats are written as `'NaN'`/`'Infinity'`/`'-Infinity'` so a strict JSON parser accepts the line.
- A log call no longer raises on content JSON refuses; the console sink costs at most the line Rich refuses; and `TracebackMessageError` renders its message verbatim.
- The `otlp` sink filters out its own export-path records before the lock, keeps mixed-type sequences as JSON text, and `Log.reset` reports flush/close failures on stderr without escaping.
- The OpenTelemetry dependencies now declare the minimum versions the `otlp` sink needs, so its keyword-only `emit` call works at the declared floor.

<sup>Written for commit a598015e9f49b75b0b0ab6b9145b817d19fc1530. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Pipelex/pipelex/pull/1206?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lchoquel:
Review pass, round 2 at profile 4, bar defects. Reviewers that ran: cubic, codex:review, codex:adversarial, code-review; one verifier reproduced the fix group.

Fixed:
- 78706f6c7 — the sink handoff delivers a record emitted during the install exactly once, after the held ones (the forwarded record is marked and the sink's guard rejects the root's second delivery).
- 47df92136 — a flush that raises no longer skips the sink's close, and the doctor releases the logging it configured once its report is out.
- f3ba8e7f4 — the doctor reports a plugin that stops discovery and a sink that fails to install as rows of their own, on a console sink pinned to stderr; the stale `rich_handler` docstring in the agent CLI factory is corrected.
- 67ba55a24 — the silencing recipe the log-mode retirement names is the package level under `[runtime.log.package_log_levels]`, the one that governs Pipelex's loggers.

Deferred, traced in `wip/structured-logs/review-deferrals.md` (round 2 section): the width of the structured-content guard in the dispatch, the doctor keeping an already-configured sink without resolving its token, and the safe migration's untranslated keys restated from round 1. Rejected: the concurrent `install_sink` race, since the boot configures logging once per process and refuses a second call.

lchoquel:
Review pass, round 3 at profile 4, bar necessity. Reviewers that ran: cubic, codex:review, codex:adversarial, code-review; the two admitted findings were reproduced by the driver.

Fixed:
- fa0cfa0fd — the holding handler delivers a replayed or forwarded record the way the root logger would: it checks the sink handler's level, which `Handler.handle` does not, and routes a record the handler cannot render to `handleError` instead of raising out of the log call.
- 0a4177870 — a teardown diagnostic that cannot be written no longer stops the reset; the configuration is forgotten in a `finally`, so the next configure is not refused.

Deferred, traced in `wip/structured-logs/review-deferrals.md` (round 3 section): a sink dropped without teardown when its handler build fails after allocating resources, the safe migration's untranslated keys at boot recovery (the boot warning names the by-hand edit), the doctor's second composition root, the unguarded processor loop (handed to the scrub member, which registers the first processor), the OTLP sink overwriting a carried `code.*` field, and two nits.

lchoquel:
Review pass, round 4 at profile 4, bar freeze: this pass changed no code. Reviewers that ran: cubic, codex:review, codex:adversarial, code-review. Nothing found meets the bar's critical definition.

Two findings were reproduced by the driver and filed as ledger bugs rather than fixed here, each a one-line fix with a test: a caller field spelled `_pipelex_forwarded` is read as the handoff mark and the record is dropped by the sink (L-260916-40bf74); the doctor's fallback sink install raises "already installed" when the configured sink failed after the replay recorded it (L-260916-7e131d).

Deferred, traced in `wip/structured-logs/review-deferrals.md` (round 4 section): a caller's `data` field overwritten by structured content, sink factories taking the whole log config, processors mutating the record every root handler shares (carried to the scrub member), held records pinning live tracebacks, and a recursion nit in the JSON rendering. Restated from earlier rounds and standing as traced: the safe migration of a silenced file, the OTLP close order, the linear drop at capacity. Refuted again: the kit configs without an `[runtime.log.otlp]` table.

lchoquel:
**Review round 1 at profile 4, bar `open`** — the branch grew past the growth line after round 4, so the ladder reset and all four reviewers (cubic, `codex:review`, `codex:adversarial`, `code-review`) read the whole branch against `origin/dev` again. Twelve findings; eight verified against the branch head, four sorted as standing deferrals already traced.

Worth recording first: **none of the twelve is about the lines whose arrival reset the ladder** — the forwarding-marker reservation and the OTLP reserved keys are the part of this branch the rounds have now read most and found least in.

**Five fixed**, each mutation-tested by reverting the fix and watching its own guard go red with nothing else moving:

- A non-finite float nested in a pydantic model passed in `fields=` made `json.dumps` refuse the payload under `allow_nan=False`, and the line's own recovery reprs every value it carries — one `NaN` turned an integer field into a string and a list into its `repr`. `json_fallback` now spells a model's dump the way a value handed in directly is spelled, which fixes the `otlp` sink's `render_json` on the same route. Reachable with `sink = "json"`, which the hosted plane selects.
- `release_to` drained held records without marking them, where the forward path marks, so a thread descheduled inside `callHandlers` before the handoff could reach the sink's handler with a record the drain had already delivered. A distinct window from the one the forwarding-marker commit closed, on the other side of the drain.
- `HoldingLogHandler.close` dropped every held record below the last resort's own WARNING, so a boot that died between `configure` and `install_sink` lost exactly the trail a verbose run was turned on to produce. That level is meant for a record emitted before anything was configured; these passed the level the configuration set.
- A field spelling `data` beside structured content was destroyed rather than carried under `field_data` — the one collision in the module that broke the discipline the module states everywhere else. Carried as a round-4 deferral, taken here.
- The four `opentelemetry-*` dependencies carried no version specifier at all, and no job resolves the floor. The SDK took a positional `emit(record)` in versions the declaration permitted while the sink passes the keyword form, so such a resolution made every record a `TypeError` that `handleError` turned into a stderr traceback while the process reported success. Floors declared at the locked versions.

**Four deferred**, each with a trace in `pipelex/wip/structured-logs/review-deferrals.md`:

- A processor that raises kills the live log call, where the replay path swallows the same exception — stays with `L-260916-4837d4`, which registers the first processor and owns the contract choice (a redaction processor that raises means the record is *unredacted*, so rendering it anyway leaks what it was meant to hide).
- The `otlp` export-path filter rejects every record emitted during any export, not only the export path's own; the concrete loss is Pipelex's span-export diagnostics. Filed as `L-260917-472a3e`. Not narrowed here because the guard's breadth is what fixed a reproduced exit deadlock.
- A boot that fails inside `__init__` after `configure` never closes the holding handler at all, so nothing it held is emitted anywhere. Filed as `L-260917-faebf7`.
- The doctor's fallback install (`L-260916-7e131d`) was confirmed, and the verification moved where its fix belongs: the ordering in `install_sink` is deliberate — the sink is recorded before the replay so `reset` can remove a handler already on the root logger — so the cure is in `doctor_cmd.py`.

Plus four standing deferrals re-sorted unchanged, all with traces already in place: processors mutating a shared record, the doctor duplicating the boot policy, external sinks having no plugin-owned configuration boundary, and `CALLER_FRAME_DEPTH`.

`make agent-check` passes in full and `make agent-test` reports all tests passed at the new tip.

lchoquel:
**Review round 2 at profile 4, bar `defects`** — same four reviewers (cubic, `codex:review`, `codex:adversarial`, `code-review`) over the branch at `7490b838b`, with a focus naming the three commits round 1 added.

Three of the four found only findings already deferred with traces: cubic four, `codex:review` two, `codex:adversarial` one. `code-review` found new ground, and one of its findings is the most serious thing six rounds on this branch have turned up.

**A log call could raise on the default sink — the one thing this branch's changelog promises it never does.**

It arrived as a claim about one error path. Verifying it made it larger and moved its cause. `RichHandler` overrides `emit` and does not restore the `try`/`handleError` the stdlib's own handlers put around theirs, so anything raised while rendering leaves the log call — and Rich's markup parser raises on ordinary text. Its tag must open with a lowercase letter, `#`, `/` or `@`, which is where the finding's own example was wrong: `[Errno 2]` is safe. The two shapes an error message is actually made of are not:

| message | before |
|---|---|
| `[Errno 2] No such file or directory` | intact — uppercase is not a tag |
| `expected list[int], got str` | the bracketed span silently lost |
| `no such file: [/etc/pipelex.toml]` | **raises `MarkupError` out of the log call** |

Confirmed end to end against the branch before the fix: `log.error("no such file: [/etc/pipelex.toml]")` raised, and constructing a `FatalError` carrying such a message raised `MarkupError` **in place of** the failure it was reporting — the diagnostic destroyed and replaced by an unrelated exception.

Fixed at both levels. The sink's handler restores the stdlib's guard, so a line Rich refuses costs at most its own rendering. And `TracebackMessageError` asks for its record verbatim, through the per-record override Rich reads ahead of its handler's setting — reserved in `log_fields` beside the forwarding marker, for the same two reasons: it is a console concern that rides no wire, and a caller's field spelling it must not decide how the console reads that line. Both arms of the error mode are marked, the `error` one this branch moved onto a propagating logger and the `exception` one `dev` already exposed.

**A guard this branch deleted is restored in a form that survives what retired it.** Two tests pinned that the gateway prune runs before the log hub is configured, by asserting an unconfigured dispatch raised; a log call before `configure` is now handled rather than fatal, so they stopped failing for their own reason and were removed — while `drop_unknown_gateway_defaults` goes on stating "deliberately pure, and deliberately silent" in its docstring with nothing holding it to it. The `defaults` guard now asserts the silence directly. Its per-model twin is deliberately not restored: nothing in that path's source claims silence, so pinning it would invent a contract rather than hold one.

**Deferred and filed:** `L-260917-0be44e`, the doctor emitting its own diagnostics through the configured sink, so with `sink = "otlp"` its stale-configuration warning never reaches the person running it — deferred because the remedy is contested inside this same round, between that finding and cubic's standing one that the doctor already duplicates too much of `RuntimeBoot.setup`. And `L-260917-398eae`, the `json` sink carrying ANSI escapes from a third-party record decoration, which is the hosted plane's own configuration; unverified, since no uvicorn is installed here.

**Rejected, with the reason recorded so neither returns as new:** the `_render_content` re-parse claim (the state cannot occur — the rendering is `json.dumps` output of an already-purified structure, and truncation runs on the structure and not on its text), and the holding handler's overflow order (a contested preference, on a path a real boot does not reach, against a constant whose comment states the choice and its reason).

Full traces in `pipelex/wip/structured-logs/review-deferrals.md`. Each fix was mutation-tested — reverted, its own guard watched going red, nothing else moving. `make agent-check` passes in full and `make agent-test` reports all tests passed at the new tip.

lchoquel:
**Review round 3 — profile 4, bar `necessity`, outcome `clean`.** All four reviewers ran over the whole branch against `origin/dev` at `2e81b2c68`: cubic, `codex:review`, `codex:adversarial` and `code-review` at level medium. No code changed. Nine findings merged; nothing the bar admits was confirmed.

**The processor seam came back for the fourth consecutive round, this time from two reviewers.** `codex:adversarial` at confidence 1.0 and `codex:review` at P2 both name `pipelex/tools/log/log_sink.py:60-63`: `_ProcessorFilter.filter` runs its callbacks where the stdlib runs a handler's filters, outside every `try`, so a processor raising on a live record raises out of `log.info` itself, while the same processor raising during the holding replay is contained by `_deliver`'s `handleError`. It stays deferred for the reason the earlier rounds recorded — the remedy is the scrub member's contract choice, since a redaction processor that raises means the record is *unredacted* and rendering it anyway leaks what it was meant to hide — and no processor is registered anywhere in `pipelex/` today. The repetition is worth saying out loud: whoever merges this should decide it deliberately rather than let it default through.

**The one new claim the bar would have admitted was verified and refuted on attribution.** cubic raised at P1 that `configure()` overwrites the root level and every configured package logger's level while `reset()` restores none of them, so an embedder's logging stays changed after `teardown()`. The mechanism is real and now carries a verified failing scenario — but `origin/dev` overwrites the same two families plus a third, restores nothing either, and strips the host's own root handlers on top, so this branch strictly shrinks the residue. The new doctor `finally: log.reset()` sites are guarded by `if not logging_was_configured` and do not widen the exposure. Filed as `L-260918-aee78b` with the full verification.

Also deferred, all new this round and all unverified: `pipelex doctor` can block a minute or more flushing the `otlp` sink when the collector is down; `stream_for_target`'s docstring promises a lazily-read stream no caller reads lazily; a `LogSink` instance is single-use because `reset()` closes the handler it caches, and nothing says so; and a sink whose factory already allocated a thread and an `atexit` hook is dropped undisposed when `install_sink` raises. `codex:review` re-raised the OTLP suppression guard with the PostHog mechanism already on record (`L-260917-472a3e`).

Rejected: the holding handler's oldest-first overflow, on round 2's argument with nothing new offered; and the fixed caller frame depth, which is the fields member's code and describes a change nobody has made.

Traces are in `wip/structured-logs/review-deferrals.md` under the round-3 heading, pushed as b15446950.

lchoquel:
**After round 3: one fix landed and one deferral became a decision.** Neither came out of the round's own findings — both are the owner's answers to what round 3 left standing.

**The processor seam is settled onto the scrub member, `L-260916-4837d4`, rather than deferred a fifth time.** Four rounds and two independent reviewers had agreed on the mechanism, and the venue was never in doubt: a redaction processor that raises means the record is unredacted, so the remedy belongs where a processor first exists. What was missing was the obligation, and the scrub member now carries it — decide the fail-closed policy and say why in the code, guard the boundary into `handleError`, and cover the live path *and* the holding replay, which behave differently today with only one of them tested. The trace is in `wip/structured-logs/review-deferrals.md` under the processor-seam heading, `9031bb421`.

**`a06282ecf` closes `L-260916-7e131d`: the doctor's console fallback no longer installs over a sink already recorded.** `log.install_sink` records the sink before it replays what the holding handler held — deliberately, so a replay that raises still leaves the sink findable by `reset` — so a failed installation does not always leave nothing behind. Installing on top raised `RuntimeError: A log sink is already installed` in place of the failure the doctor was called to report, and the row promised a console fallback that never stood in. `_install_fallback_log_sink` now decides whether it can stand in and returns the note the row carries; the three arms interpolate that note. Test-first, failing with exactly that `RuntimeError` before the guard, then mutation-tested by handing back the normal note in the refusal arm — one test red, no other moved. The `cli-docs` drift contract opened and was paid: `pipelex/cli/agent_cli/CLAUDE.md` had promised a not-healthy `log_sink` row means the report ran on the console sink, which is the claim this qualifies.

**The wider version of the same finding was verified and deliberately not taken.** The doctor reporting its `log_sink` row healthy when logging was configured but no sink was ever installed is real as a mechanism; the scenario and consequence as first reported were wrong, the severity is low, and the clean fix is a discriminant carried through every row that reads it, about ten files. Filed as `L-260918-d5355f` with what makes it worth doing: `setup_doctor_runtime` owns that arm and no test calls it — every reference under `tests/` patches it wholesale, which is why this survived six rounds. The same pass found two tests in `test_doctor_log_sink.py` that were order-dependent, patching `log.install_sink` while leaving `log.sink` real, and both are now on the `released_log` fixture.

`make agent-check` and `make agent-test` are green on the branch. This commit changes code, so the ladder reopens — read the next bar from `ledger review-profile`, not from the round number.

PR #1207, opened 2026-09-16
feature/Log-redaction · L-260916-4837d4 — scrub secrets, safe values

lchoquel:
Adds a redaction processor that every installed sink runs first: secret-bearing text is scrubbed by pattern and by field name, structured values are rendered to safe primitives before the walk, exception text is scrubbed into `exc_text`, and a redaction failure quarantines the record rather than letting it through. It is on by default under `[runtime.log.redaction]` with `extra_patterns` for a deployment's own shapes. The branch also reserves the forwarded mark from field collisions and makes the doctor fallback fire only when no sink is installed.

Closes L-260916-4837d4
Closes L-260916-40bf74
Closes L-260916-7e131d
Closes L-260916-715f62
Advances L-260916-8bcd3f

https://claude.ai/code/session_016gUk6TEF3BG4qc4qeaSfvy

---
## Summary by cubic
Scrubs secrets out of every log record before any sink renders it, configurable under `[runtime.log.redaction]` and on by default, and fixes several log-handling bugs.

**New Features**
- Replaces bearer tokens, api-key and signature headers, OAuth codes, cookies, JSON secret entries, and `sk_` / `plx_sk_` / `pk_` / `bl_` key prefixes with `[REDACTED]` in the message, exception text, and every field string, each family reading the shape a secret is actually written in (quoted cookie crumbs, percent-encoded OAuth codes, `Basic` credentials).
- A field named like a secret loses its value, judged by the name the call used; structured content is scrubbed by name before the dispatch renders it, including content `json` refuses.
- Renders structured values to safe primitives before the walk, cuts cycles with a `[cycle]` marker, and neutralises control characters in field strings.
- If a scrub cannot finish, the record is dropped rather than emitted with what the scrub never read; an unrenderable value becomes an `[UNRENDERABLE: …]` marker and bad format arguments an `[ARGUMENTS WITHHELD]` notice.
- Exception text is scrubbed into `exc_text`; the `otlp` sink emits `exception.type` and `exception.stacktrace`, no `exception.message`, including traceback-less exceptions.

**Bug Fixes**
- A sink processor that raises goes to the processor guard's own diagnostic instead of raising out of the log call.
- A field named `_pipelex_forwarded` is carried under the `field_` prefix instead of rejecting the record from every sink.
- `pipelex doctor` falls back to the console sink only when nothing is installed; a sink that failed during replay is reported and kept.
- `install_sink` removes its processor from a sink whose handler fails to build, and a sink re-installed after a reset gets a fresh handler instead of the teardown's closed one.
- An `extra_patterns` regex matching the empty string is refused at load.

Closes L-260916-4837d4, L-260916-40bf74, L-260916-7e131d, L-260916-715f62, L-260916-8bcd3f.

<sup>Written for commit ee53a29d450ec5db4a39de2f6452e1cb196ec5df. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Pipelex/pipelex/pull/1207?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lchoquel:
**Review round 2 — profile 4, bar `defects`.** Reviewers: cubic, Codex (review and adversarial), code-review at `medium`, each on `origin/feature/Log-sink-seam..7e8f5b7`. Ten distinct findings after merging; one verifier batch confirmed eight, refuted one as unreachable and read one as a documented limit.

Fixed in 2e5ce988a:

- A field of the record whose own name is a secret's was walked by value only, so `fields={"password": "…"}` and a third-party `extra=` reached the `json` and `otlp` sinks in clear; it is now redacted by name, exactly as a mapping entry is.
- The message of a structured content is rendered from the redacted content, so it no longer carries what `data` lost — a secret nested in an object, or held as a number, is beyond what a pattern reads back out of text.
- The serialised-entry family reads its names in any case, with a dash or an underscore, against a value quoted either way and against a bare number or boolean.
- The processor guard reported a failure through the stdlib's `handleError`, which printed the exception that had just failed on the record's values; it now names the processor and the exception type and quotes nothing.
- A record whose arguments do not fit its format string was left for the handler to fail on, and the stdlib printed every argument raw; it keeps the scrubbed format string and an `[ARGUMENTS WITHHELD: …]` notice.
- One value that refuses to render (a model whose dump raises, an object whose `__str__` raises) quarantined the whole record; it becomes an `[UNRENDERABLE: …]` marker and the rest of the line survives.
- `install_sink` left its processor on a sink whose handler build raised, which `reset` could never reach, so the same sink installed again carried two.
- The `otlp` sink exported only `exception.type` for an exception carrying no traceback, losing its text.

Deferred, with their traces in `wip/structured-logs/review-deferrals.md`: a sink's own processor running after the redaction (refuted as reachable — no sink in the workspace appends one); the `console` sink's Rich traceback rendering from the exception object (a documented limit, carried by its own ledger item); a mapping key that is itself a secret; a control character in a mapping key; the `otlp` sink exporting no exception for a record rebuilt from another process (pre-existing); and a secret nested in text a call rendered itself.

`make agent-check` and `make agent-test` are green at 2e5ce988a.

lchoquel:
**Review round 4 — profile 4, bar `necessity`.** Reviewers: cubic, Codex (review and adversarial), code-review at `medium`, each on `origin/feature/Log-sink-seam..2e5ce988a`. Round 3 had reported its findings without applying them, which holds this round at the same bar rather than letting the ladder climb over work nobody had done — so this round carried them, and its own reviewers raised most of them again. Eleven findings, one verifier batch, eleven confirmed by reproduction and all eleven fixed: eight a way a secret left the process, three a way the redaction destroyed what it was meant to leave alone, and four of them introduced by round 2's own fixes.

Fixed in f522d0093:

- The cookie family, both of its failure modes at once. A crumb holding a quote leaked the rest of the header (`Cookie: session="…"; HttpOnly`), and the same pattern ran to the end of any line whose prose merely said "cookie:" — destroying the sentence, and a secret a later word carried. It now reads a header's own `name=value` crumb grammar, which is what separates a header from a sentence.
- A field carried under the `field_` collision prefix escaped the name check round 2 added, so a record factory that stamps `password` had the caller's field exported in clear. The name is judged as the call spelled it.
- The OAuth `code` class stopped at `%`, so a percent-encoded code passed through untouched whenever the run before the escape was shorter than the minimum length. It is read to the delimiter that ends the parameter.
- The key-prefix family missed the dash spellings actually issued (`sk-proj-…`, `sk-ant-api03-…`), matched ordinary snake_case identifiers (`pk_customer_reference_index`, `bl_run_history_partition`) and dropped the prefix that named the secret instead of keeping it.
- The secret names were read only when something had quoted them — the serialised shape, not the shape a raw request arrives in — so `client_secret=…` in a form body and `access_token=…` in a query string went through. The plain `name=value` form is a family now.
- `Authorization: Basic` and `Authorization: Token` were not scrubbed, the literal word `bearer` being required; Basic decodes to a credential pair in one step.
- `extra_patterns` accepted a pattern matching the empty string: `x*` where `x+` was meant interleaves the redaction text between every two characters of every line for the life of the process.
- A caller's `fields={"data": …}` landed on the attribute the runtime exempts from control-character escaping as its own rendering. `data` is the runtime's name outright now, and a caller's is carried as `field_data`.
- Structured content was redacted by name only where the serialization succeeded, so content `json` refuses fell back to its `repr` with an object-valued secret intact — the one rendering the string families cannot read.
- The quarantine could itself fail, at the stack depth that is the commonest reason the scrub failed at all, and hand a sink a record neither scrubbed nor stripped under a message asserting it had been quarantined. Such a record is dropped.
- A sink installed again after a `reset` was handed back the handler its teardown had closed, which re-installs the `otlp` sink dead: accepting every record, running every filter, exporting none, with nothing raised to say so.

Deferred, with their traces in `wip/structured-logs/review-deferrals.md`, all unverified: the `console` sink's Rich traceback (its own ledger item); a mapping key that is itself a secret and a control character in one; a plugin that reassigns `sink.processors` after the handler is built; `[cycle]`, the one marker Rich deletes; the raw format string a withheld-arguments line keeps; the doctor inferring a failed install's cause from a global; a record held when the boot dies before `install_sink`; and one OTLP test that does not isolate the branch it names.

`make agent-check` and `make agent-test` are green at f522d0093.

PR #1208, opened 2026-09-17
fix/Dead-natives-version-constant · L-260917-87199b

lchoquel:
`PINNED_NATIVES_MTHDS_VERSION` read `1.0.0` while the MTHDS standard pins the native set at `2.0.0`, and nothing in the package read the constant, so no check here could say it had gone stale across the standard's 2.0.0 cut — a downstream port reporting it as this reference's natives version gated its own goldens against a dead number. It now reads `2.0.0`.

It also has readers now. The standard-conformance suite compares it against the page's own "Pinned at MTHDS" statements, which is the check that would have caught the drift; a second module holds it to the standard version the installed `mthds` reports, which needs no sibling checkout and so runs for every contributor.

Closes L-260917-87199b

https://claude.ai/code/session_01FAEkzXpY71NwD3hJPqLe6X

---
## Summary by cubic
Fixes `PINNED_NATIVES_MTHDS_VERSION` so it names the set it labels, and gives it readers so the value can't go stale again.

The constant read `1.0.0` while the MTHDS standard pins the native set at `2.0.0`, and nothing read it, so no test caught the drift. It now reads `2.0.0`, and two checks keep it honest: the standard-conformance suite holds it to the page's "Pinned at MTHDS" statement, and a checkout-free check refuses a label later than the standard version the installed `mthds` implements.

Closes L-260917-87199b.

<sup>Written for commit 7f6dd229549115c3b3326fd1f55ff993f898256b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Pipelex/pipelex/pull/1208?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lchoquel:
**`/rev` round 2 — profile 2, bar `defects`.**

Reviewers that produced a review: cubic, Codex (`codex:review`, gpt-5.6-sol, effort medium) and `code-review` at level low. No missing coverage. cubic and Codex were both clean this round.

Two confirmed findings, both fixed in 7f6dd22 — factual errors in prose rather than runtime defects. The new module's docstring offered the `1.0.0` staleness as the incident it prevents, but its bound is one-directional and holds on exactly that value, so the docstring now says which direction each of the two test modules covers. The comment above the constant still described the pre-split arrangement and sent a reader to the wrong module for the engine-version guard.

Two findings deferred, both unverified, traced in `wip/pinned-natives-version/deferred-review-findings.md`: the page-version regex matches "pinned at MTHDS" inside "re-pinned" and "previously pinned", so a future page wording could trip its unanimity check; and the changelog entry narrates the test coverage it adds.

Round 1 (profile 1, bar `open`) fixed one cubic finding in 0ea3098 — the page-free guard moved into its own module so each module holds one TestClass.

PR #1210, opened 2026-09-18
feature/Rich-behind-extra · L-260916-fe7969 — rich extra, lazy import, guard

lchoquel:
Rich leaves pipelex's core dependencies for a `cli` extra: nothing outside `pipelex/cli/` imports it at module level, and the console sink, `get_console`, the pretty-print engine and every `rendered_pretty` method import it inside the body behind a guard that names the extra, with `runtime_boot` refusing `pretty_print_mode = "rich"` when it is absent. The property is measured rather than asserted — `make check-rich-imports` refuses both a module-level import and a transitive reach through a CLI module, and an integration test runs a compose pipe in a subprocess whose import hook raises on `rich`. What this deliberately does not claim is a Rich-free install: `typer` and `instructor` both require Rich unconditionally, so what ships is the narrower property that is true and measured — no pipelex code path on the way to running a method needs Rich.

This is stacked on `feature/Log-sink-seam` and targets it rather than `dev`.

Closes L-260916-fe7969

---
## Summary by cubic
Makes Rich the `cli` extra and moves every Rich import outside `pipelex/cli/` into the code that renders, so no runtime code path to running a method needs Rich.

**Migration**
- Install `pipelex[cli]` for the `pipelex` and `pipelex-agent` commands.
- Without the extra, a boot selecting the `console` sink or `rich` pretty-print mode stops naming the extra and the Rich-free alternatives; `get_console()`, `rendered_pretty()`, and Rich-table listings raise `MissingDependencyError`.
- `typer` and `instructor` still require Rich, so the environment is not Rich-free; only the runtime's path to running a method is.

**Enforcement**
- `make check-rich-imports` refuses module-level Rich imports, reaches through CLI modules, and deferred imports reached without a preceding guard; it runs in CI, `agent-check`, and the check aggregate, with `hub-layering: ignore` as its only escape hatch.
- The cost report writes its CSV before rendering the console table, and the pytest plugin's CI diagnostics no longer print through Rich, so neither output is lost without the extra.
- The `poor` mode's Rich-free renderer parses panel titles by nesting depth, so alias-paired markup like `[b]x[/bold]` renders plain, and the url frame keeps its inner title.
- An integration test blocks Rich in a subprocess and runs a compose pipe end to end; it declares no model specs, so boot needs no inference setup.

Closes L-260916-fe7969

<sup>Written for commit 3adf74560f631e7280e9416393ae4689569b13c7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Pipelex/pipelex/pull/1210?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lchoquel:
**Review round 2 — profile 4, bar `defects`, outcome `clean`.** All four reviewers ran against `origin/dev` at `1ec05b2db`: cubic (focused on this member's own commits), `codex:review`, `codex:adversarial` and `code-review` at level medium. No code changed. Nine findings merged; nothing the bar admits was confirmed.

**This branch is nine commits behind its own base, and that shaped most of the round.** `feature/Log-sink-seam` is at `b15446950`; this branch sits on `e6769a0e8`. Two of `codex:review`'s three findings are artefacts of that — the unreserved `_pipelex_forwarded` marker and the OTLP `code.*`/`exception.*` keys overwriting carried fields were both cured below at `208f2ad63`. They are not defects of the merged result. The staleness is worth clearing before this merges.

**Two findings in this member's own code were verified, and both are deferred.**

The rich-import guard's ordering rule (`rich_import_guard.py`) matches on syntactic mention, so `if not is_rich_installed(): pass` really does mark the rest of the block guarded — confirmed by running the guard over it. Deferred because it is deliberate and documented (`_establishes_rich_is_available`'s own docstring says it settles that the question was asked, not that every path through the answer is sound), because the compensating control is the Rich-refused subprocess run in `tests/integration/pipelex/test_rich_free_run.py`, and because an audit of every deferred Rich import outside `pipelex/cli/` found all of them genuinely guarded — including the two sites using exactly the shapes cited, both of which `return` in the branch. Making the rule sound needs dataflow, which is a different guard.

The doctor's fallback sink builds a `ConsoleLogSink` whose `make_handler` calls `require_rich`, so on a Rich-free install every failure arm of `install_doctor_log_sink` raises `MissingDependencyError`, which is a sibling of `PipelexConfigError` and escapes to the broad handler — `Unexpected error:` and exit 1, no report. Reproduced. Deferred because it is unreachable: `doctor_cmd.py:13-16` imports Rich at module level, legally, so with Rich blocked the command dies at import first; and it is not made live by `L-260917-c29954` alone, which makes a Rich-free install producible without making the doctor Rich-free. If the fallback is hardened anyway, the sound move is one pass swapping in `JsonLogSink(stream=sys.stderr)` **and** adding the `log.sink is None` gate `L-260916-7e131d` asks for.

**Routed to the members below:** the doctor reporting its log-sink row healthy when logging was configured but no sink was ever installed (`doctor_cmd.py:1235`, noted on `L-260916-aa6e58` — worth deciding before that member merges); and the second serialization per structured record (`log_dispatch.py:176`), an improvement this bar excludes, already traced on `L-260916-a1dbd5`.

`codex:adversarial`'s sole finding is the migration question settled by the owner on 2026-09-17. It reproduces a mechanism nobody disputed and offers no new fact, and the settlement is explicit that reopening needs a configuration that actually sets one of the retired keys, not another reading.

Traces are in `wip/structured-logs/review-deferrals.md` under the round-2 heading, pushed as 107d1ada5.

PR #1214, opened 2026-09-19
feature/Judgment-family · L-260919-9965d0 — judgment family, TypeSafe Jev

lchoquel:
The judgment-family design was written entirely from TypeSafe's documentation, so phase 1 of the campaign replaces reading with observation before any worker code exists. Two throwaway spike scripts call the live API across the three question types, the state value types, the criteria edge cases and limits, model pinning, errors, retries, concurrency and stability, and save every successful response as a wire body for phase 2's unit tests to replay. The findings note ends in a go and settles the design's open question on the vendor dependency in favour of the SDK behind a `typesafe` optional extra, which added five packages and moved nothing else in the lock.

Observation contradicted the documentation in several places and the design is corrected in the same change: validation failures are 400 rather than 422 and carry an error body of two shapes, a rating is capped at ten levels, an unknown yes/no criteria key is ignored silently, the models endpoint lists only aliases, a rating answer's probability keys are strings on the wire, usage is per request with output tokens free, and an array state degrades answers. Checkpoint 1 records the go, the dependency decision, the margin the live tests must use, and the condition attached to keeping the extra — no release until the integration is real.

Closes L-260919-3aeab8
Advances L-260919-9965d0

https://claude.ai/code/session_01SuyzidXG43oVLK5USiJMuv

---
## Summary by cubic
Replaces reading TypeSafe's docs with live observation from two spike scripts, so the judgment-family design reflects what the API actually does and phase 2 can replay saved wire bodies instead of calling the API. The findings settle the vendor dependency on the SDK behind a `typesafe` optional extra and correct the design where the docs contradicted behavior.

**Findings**
- Validation failures are 400s with two error body shapes, not 422s.
- A rating is capped at ten levels.
- An unknown yes/no criteria key is ignored silently.
- The models endpoint lists only aliases.
- A rating answer's probability keys are strings on the wire.
- Usage is per request and output tokens are free.
- An array state degrades answers.

**Phase 2**
- The plan now states the four findings the worker cannot be written without: error classification branches on `error_type`, request ids need different accessors on responses and errors, rating answer keys change type across three layers, and the deck's versioned handle is not confirmable through the models endpoint.
- Phase 2 lands as two pull requests so a reviewer can separate wiring the family in from reading TypeSafe correctly.

Adding the extra moved nothing else in the lock. Checkpoint 1 keeps the `typesafe` extra on the condition that no release goes out until the integration is real. Closes L-260919-3aeab8. Advances L-260919-9965d0 and L-260919-502f36.

<sup>Written for commit 4c823c73f57c7a5b2e2bf0c138fe7e3eb5294522. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/Pipelex/pipelex/pull/1214?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lchoquel:
**Review pass recorded** — profile 4, round 1, bar `open`, on `6cd7740f5`.

cubic and the official code-review (level `medium`) both ran and both showed provenance matching the plan. **Codex did not run**: the ChatGPT usage limit was hit, so this pass is one reviewer short and the record says so. Thirteen merged claims went to a verifier: twelve confirmed, one refuted (cubic's claim that the plan's `uv run` command cannot work — `--no-project` still discovers the worktree's `.venv`, where `httpx` is a core dependency).

Fixed in `cf21205bb`: the live key's length removed from a committed fixture; the ten-level cap probed at its boundary (ten accepted, eleven refused) instead of inferred from a refusal's wording after sending 24; `raw_http_response` and `request_id` guarded correctly, since both raise rather than returning `None` — the worker needs two different accessors and the findings note now records that; the follow-up fixtures switched from the SDK's `model_dump` to wire bodies, with the five silent probes now saving; the usage series replaced with a single experiment; the ten-level cap assigned to the worker, resolving the design against the findings note and the plan; and four smaller spike defects.

One decision was taken by the developer rather than by the review: the `typesafe` extra ships without a changelog entry, on the condition that no release goes out until the TypeSafe integration is actually in. Checkpoint 1 records it.

